PR #25, opened 2020-04-04
Bryant integrate icp seg

smguo:


bryantChhun:
Mostly minor changes: changes to OD calculation.  Commented out some debug statements.  And notebook updates

PR #26, opened 2020-04-05
tested the analysis pipeline with flu plate data.

mattersoflight:


bryantChhun:
@mattersoflight can you add a description of what this PR does?  I see a lot of notebooks but the names are not clear to me (such as "interpretationV1, V2, V3, V4").  Perhaps you can add a line in the description above about which notebooks are important?

smguo (approved):
I tested both the interpolation and fit options in the PR. Both run fine with the same issues that @Shalin Mehta and @Bryant already documented in Issues

mattersoflight:
Hi @bryantChhun notebook versions v1, v2, are attempts at reading the output from pipeline and plotting with pandas/seaborn. Won't affect the array_analyzer. The lasting edits we need from here are in the workflow, which provide improvements to command line and debug plots in interpolation workflow.

PR #32, opened 2020-04-07
Bryant od calc in icp

bryantChhun:
@jennyfolkesson 
A few small changes:
- calling "get_spot_coords" and "create_reference_grid" from the appropriate location
- fixed "build_centroid_binary_blocks" to receive "crop_coords" not "reg_coords"
- fixed background plotting scale factor
- during plotting, shifted the spot/grid coord to match the crop_coords


jennyfolkesson (approved):
Thanks for fixing function calls - I was a bit tired last night when I made this branch. :-)

PR #34, opened 2020-04-08
Robustspotdetection

jennyfolkesson:
I believe this should be stable now, but @bryantChhun should have a final say before this is merged. I wanted to start the code review at least.

One question - it looks like a lot of images are displayed now as you're running, and that makes the run pretty slow, plus you have all these plots popping up. What about changing this to write only?


mattersoflight:
I suggest @smguo merge his recent changes that improve the estimation of OD into this branch, and then we test it against March 25 and Apr 4 datasets. 

@jennyfolkesson these plots are to catch edge cases as the pipeline is developed, but they are all invoked if debug is enabled. If there are any plots that are outside if debug: flag, we should move them there.

bryantChhun:
There is a minor issue with the OD calculation -- both the cropped image and the background need to be on range 0-1, float64, not inverted

I will submit a PR to robostspotdetection

jennyfolkesson:
@mattersoflight @bryantChhun @smguo I believe this PR is ready for review now.

run_array_analyzer.py
```diff
@@ -1,80 +1,85 @@
...
-
-    if not os.path.isdir(inputfolder):
-        raise ValueError("input folder is not a folder or not supplied")
-
-    if not os.path.isdir(outputfolder):
-        os.makedirs(outputfolder)
+import argparse
+import os
+
+import array_analyzer.workflows.registration_wf as registration_wf
```
smguo: ModuleNotFoundError: No module named 'array_analyzer.workflows.registration_wf'

array_analyzer/workflows/registration_workflow.py
```diff
@@ -0,0 +1,272 @@
...
+                im_crop,
+                props_array_placed,
+                well_path,
+                image_name[:-4],
+                from_source=True,
+            )
+            print(f"Time to save debug images: {time.time()-start_time} s")
+
+            # # Save image with spots
+            im_roi = (np.iinfo(im_crop.dtype).max*im_crop.copy()).astype('uint8')
```
smguo: Got this: ValueError: Invalid integer data type 'f'. 
it looks like the allowed dtype is int but float is passed  

I guess this is what you were trying to do:
`im_roi = (np.iinfo(np.uint8).max*im_crop.copy()).astype('uint8')`

bryantChhun: I'm having some difficulty with this line.  Not sure how it didn't come up earlier

"im_crop" is float64 at this point, so np.iinfo throws an error.  Additionally, we are casting float to uint8 just like  earlier, so probably just need 255*im_crop

run_array_analyzer.py
bryantChhun: I believe this is now named "registration_workflow"

smguo:
I tested both workflows on the COVID_0408 plate data. The fit workflow runs fine after fixing the lines I noted in comments, but the interpolation workflow errors out on C1, which I didn't encounter in my branch. I'll submit a separate PR to fix the issue with interpolation workflow.

The fit approach seems to work well on he COVID_0408 plate data. If does fail on a few wells but interestingly the failed wells seem to complement the failed wells in interpolation workflow. I'll submit an issue to elaborate on this.   

Fit approach output:
ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/Automated_Analysis_Spot-Fitting/4_10_13_21_19

interpolation approach output (using `spot_intensity_fix` branch):
ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/Automated_Analysis_Spot-Fitting/4_9_19_25_45

bryantChhun (changes requested):

array_analyzer/workflows/registration_workflow.py
jennyfolkesson: I thought you removed all the iinfo after the first one but somehow missed this one. Will remove.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -0,0 +1,251 @@
...
+            params['columns'],
+            dtype=object,
+        )
+        bgprops_array = txt_parser.create_array(
+            params['rows'],
+            params['columns'],
+            dtype=object,
+        )
+
+        nbr_grid_rows, nbr_grid_cols = props_array.shape
```
smguo: Are `nbr_grid_rows, nbr_grid_cols` same as  `params['rows'], params['columns']`?

run_array_analyzer.py
```diff
@@ -1,80 +1,85 @@
...
+
+
+if __name__ == '__main__':
+    args = parse_args()
+    run_workflow(
+        input_dir=args.input,
+        output_dir=args.output,
+        method=args.method,
+        debug=args.debug,
+    )
```
smguo: Nice!

array_analyzer/workflows/registration_workflow.py
```diff
@@ -0,0 +1,251 @@
...
+                bg_well,
+                output_name,
+            )
+            debug_plots.save_composite_spots(
+                im_crop,
+                props_array_placed,
+                output_name,
+                from_source=True,
+            )
+
```
bryantChhun: it looks like the plots for the background images have been removed and are not present in the "debug_images" module.  Plotting background images were exactly what told me there was an issue with scaling and bitdepth.  Will this become a test instead?

bryantChhun:
I'll work on merging this into master.  There was a bit of module-movement, so I'll need to refactor a little, then retest across test data using interp/fit/well.

Will create a new branch for those changes and submit a PR here.

jennyfolkesson:
I didn't rebase robustspotdetection wrt master recently, do you want me to
do that?

bryantChhun (commented):
I'm ok to merge this PR as is

array_analyzer/workflows/registration_workflow.py
smguo: I agree with @bryantChhun. I think it would be better to keep some of the useful debug plots or move them to "debug_images" in case we need to look at them in the future. We can speed up the processing by running in regular mode instead of debug mode if the speed is the concern here.

array_analyzer/extract/img_processing.py
```diff
@@ -142,28 +155,40 @@ def get_spot_coords(im,
...
+    blob_params.minDistBetweenBlobs = min_dist_between_blobs
+    blob_params.minRepeatability = min_repeatability
+    # This detects bright spots, which they are after top hat
+    blob_params.blobColor = 255
+
+    detector = cv.SimpleBlobDetector_create(blob_params)
+
+    im_norm = cv.GaussianBlur(im, (blur_sigma, blur_sigma), 0)
+    # Black top hat filter, turns spots bright
+    im_norm = black_tophat(im_norm, size=(tophat_size, tophat_size))
```
smguo, line 190: Just curious about the purpose of adding black tophat. Is blob detector better at detecting bright v.s. dark spots?

array_analyzer/transform/point_registration.py
```diff
@@ -80,8 +80,136 @@ def icp(source, target, max_iterate=50, matrix_diff=1.):
...
+
+        for p in range(nbr_particles):
+            particle = particles[p]
+            # Generate transformation matrix
+            t_matrix = get_translation_matrix(particle)
+            trans_coords = cv.transform(np.array([fiducial_coords]), t_matrix)
+            trans_coords = trans_coords[0].astype(np.float32)
+            # Find nearest spots
+            ret, results, neighbors, dist = knn.findNearest(trans_coords, 1)
+            if remove_outlier:
```
smguo, line 179: It looks like the the outlier (fiducial that is not detected) will be removed starting at the first iteration. I'm wondering if it's possible that the wrong fiducial will be removed if there happens to be a spot close to the missing fiducial in the initial grid.

array_analyzer/transform/point_registration.py
```diff
@@ -80,8 +80,136 @@ def icp(source, target, max_iterate=50, matrix_diff=1.):
...
+                # Remove worst fitted spot
+                remove_spot = np.where(np.squeeze(dist) != np.amax(dist))[0]
```
smguo: Is this line to get the Boolean indices of the distances to keep? if so wouldn't  `dist_ids = (dist != np.amax(dist))` do the same?

array_analyzer/transform/point_registration.py
```diff
@@ -80,8 +80,136 @@ def icp(source, target, max_iterate=50, matrix_diff=1.):
...
+                dist = dist[remove_spot]
```
smguo: Is `remove_spot` actually indices of the distances to keep? Maybe change the name?

smguo:
https://github.com/czbiohub/serology-COVID19/blob/ed5783b449aae66758edf701958f2d779bc5aaed/array_analyzer/transform/point_registration.py#L31

https://github.com/czbiohub/serology-COVID19/blob/ed5783b449aae66758edf701958f2d779bc5aaed/array_analyzer/extract/image_parser.py#L310
Are `icp` and `find_fiducials_markers` just old version of `particle_filter`? Is so, can we remove them?

mattersoflight (approved):
I tested the branch with last two datasets (blinded flu plate and covid concentration test plate) for accuracy of spot detection and optical density estimation. They both look great. Ready to merge with debug plots for background added back to the pipeline.

smguo (approved):
@jennyfolkesson The PR looks great. I only have some minor suggestions. Feel free to merge when you think it's ready.

array_analyzer/workflows/registration_workflow.py
jennyfolkesson: Well, look at that! I didn't pay much attention to params I guess... :-)

array_analyzer/extract/img_processing.py
jennyfolkesson: No, the blob detector just needs a flag to know if the spots are bright or dark.
The black tophat is for detecting dark spots whereas a regular tophat would detect bright spots.

array_analyzer/transform/point_registration.py
jennyfolkesson: With enough stochasticity it doesn't seem to be a problem. All the other spots need to be in agreement with the grid so if there's one spot that's not quite on the grid that will still be the one excluded would be my guess.

array_analyzer/transform/point_registration.py
jennyfolkesson: Much better, thanks (what was I thinking...).

array_analyzer/transform/point_registration.py
jennyfolkesson: I agree, that was a bad name. I removed that variable altogether.

array_analyzer/transform/point_registration.py
smguo: I see. I guess the particles with the missing fiducial distance removed will have higher weights than the wrong solution, so these particles will eventually dominate through resampling. Very cool!

bryantChhun (commented):
...

PR #38, opened 2020-04-08
fixed and verified OD calculation

bryantChhun:
To be consistent with how OD is calculated using the "interpolation" method, we have to rescale the cropped image to be range 0-1, float64, not inverted.

After doing this, I can verify by eye that the plots from the two approaches (interpolation and fit) are similar.  This verification was done only on April flu plates.  Please see [data here](https://drive.google.com/drive/folders/1--r_pLNfQB8pT8duRBYuLAocqUAmwNDC?usp=sharing) for the debug outputs of that run:

### DIFF
--- a/array_analyzer/workflows/registration_workflow.py
+++ b/array_analyzer/workflows/registration_workflow.py
@@ -144,13 +144,13 @@ def point_registration(input_folder, output_folder, debug=False):
         )
 
         # Estimate and remove background
+        im_crop = im_crop.astype('float64')
+        im_crop /= 255.0
         background = img_processing.get_background(
             im_crop,
             fit_order=2,
-            normalize=True,
+            normalize=False,
         )
-        im_crop = im_crop / background
-        im_crop = u.invert(im_crop)
 
         placed_spotmask = array_gen.build_centroid_binary_blocks(
             crop_coords,
@@ -232,7 +232,7 @@ def point_registration(input_folder, output_folder, debug=False):
             # Evaluate accuracy of background estimation with green (image), magenta (background) overlay.
             im_bg_overlay = np.stack([background, im_crop, background], axis=2)
             io.imsave(output_name + "_crop_bg_overlay.png",
-                      im_bg_overlay.astype('uint8'))
+                      (255 * im_bg_overlay).astype('uint8'))
 
             # This plot shows which spots have been assigned what index.
             debug_plots.plot_spot_assignment(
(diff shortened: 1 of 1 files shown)


array_analyzer/workflows/registration_workflow.py
```diff
@@ -144,13 +144,13 @@ def point_registration(input_folder, output_folder, debug=False):
         )
 
         # Estimate and remove background
+        im_crop = im_crop.astype('float64')
+        im_crop /= 255.0
```
jennyfolkesson: Instead of 255 I would recommend dividing by np.iinfo(im_crop.dtype).max just in case not all images are uint8.
And you don't have to convert to a float beforehand, the division will do that automatically.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -144,13 +144,13 @@ def point_registration(input_folder, output_folder, debug=False):
...
         background = img_processing.get_background(
             im_crop,
             fit_order=2,
-            normalize=True,
+            normalize=False,
```
jennyfolkesson: False is default so you can remove that parameter when working with normalized images.

jennyfolkesson (changes requested):
Looks good, I just have some minor suggestions.

array_analyzer/workflows/registration_workflow.py
bryantChhun: Ok we'll have to change it to
```
im_crop = im_crop/np.iinfo(im_crop.dtype).max
```

The problem I had with /= 255.0 is that numpy will complain if the two data types are different.  But I always thought the /= is faster?

jennyfolkesson:
The main point is that I'd prefer to use iinfo than a hardcoded number. If
you can work that in with using /= that's your call but not sure how that
will work with different types. I think the speed difference is minor in
the grand scheme of things.

smguo (approved):
Looks good. To clarify the OD calculation: The data are absorption images so the OD should be computed by log10(background/signal). This is different from fluorescence

bryantChhun:
> Looks good, I just have some minor suggestions.

Sorry.  Missed a few more of those "255" hardcoded locations.  They are updated now.  Should be good to go.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -225,14 +223,14 @@ def point_registration(input_folder, output_folder, debug=False):
 
             # Save mask of the well, cropped grayscale image, cropped spot segmentation.
             io.imsave(output_name + "_well_mask.png",
-                      (255 * placed_spotmask).astype('uint8'))
+                      (np.iinfo(placed_spotmask.dtype).max * placed_spotmask).astype('uint8'))
```
jennyfolkesson: Actually here you still need to multiply with 255 otherwise you'll risk saving a binary image, e.g. if spotmask is a normalized float.

bryantChhun: oops yes.  I misinterpreted your "requested changes".  Will change these back.  Thanks!

jennyfolkesson (approved):

PR #44, opened 2020-04-13
Well analyzer

ieivanov:


run_array_analyzer.py
```diff
@@ -25,7 +26,14 @@ def parse_args():
         help="Output directory path",
     )
     parser.add_argument(
-        '-m', '--method',
+        '-f', '--format',
+        type=str,
+        choices=['well', 'array'],
+        default='array',
```
mattersoflight: Creating a workflow per assay format (and corresponding data format) is a good idea. It allows reuse of modules, e.g. compute_OD could be reused in this case. It also makes maintenance of each format easier. 

Instead of `--format well` and `--format array`, I suggest more obvious `--fomat elisa` and `--format scienion-array` flags.

array_analyzer/workflows/well_wf.py
```diff
@@ -0,0 +1,79 @@
...
+from array_analyzer.extract.image_parser import *
+
+import time
+from datetime import datetime
+import skimage.io as io
+import pandas as pd
+import string
+
+
+def well_analysis(input_folder_, output_folder_, debug=False):
```
mattersoflight: refactor module name and function name to be more obvious: `elisa_workflow` and this function `elisa_analysis`

array_analyzer/workflows/well_wf.py
```diff
@@ -0,0 +1,79 @@
...
+
+        well_image_dir = [file for file in os.listdir(os.path.join(input_folder_, well_dir))
+                          if '.png' in file or '.tif' in file or '.jpg' in file][0]
+        image, image_name = read_to_grey(os.path.join(input_folder_, well_dir), well_image_dir)
+        print(well_dir)
+
+        # get well intensity
+        well_mask = get_well_mask(image, segmethod='otsu')
+        int_well_ = get_well_intensity(image, well_mask)
+        int_well.append(int_well_)
```
mattersoflight: Try cropping the center of the image to simplify analysis. Also, compute (see `compute_od`) and export OD using intensities of blank wells as background.

mattersoflight (commented):
@ieivanov  In addition to following comments for converting images->OD,  please add a notebook that interprets results of this assay with inputs from @jrbyrum13  and Elias.

run_array_analyzer.py
bryantChhun: I would vote for "elisa" and "array" So that we do not have hyphens in the argument name itself.

ieivanov: The same workflow can also be used for non-ELISA measurements, so I thought that something like `well` and `array` are fitting. Thought the workflows may require some modifications between different platforms - in that case platform specific naming (like `96wp_hummingbird`, `scienion_array`, or `octopi_array`) would be more appropriate.

mattersoflight: `--format` flag  connotes antigen array format. `elisa` and `array` are good values. If we extend analysis to echo or luminex formats, we can change these values.

Single workflow should be able to handle data from different readers.  The parameters of the imagers (hummingbird, Scienion reader, octopi) are captured by the configuration metadata that @bryantChhun is working on.

array_analyzer/extract/image_parser.py
```diff
@@ -117,6 +117,29 @@ def thresh_and_binarize(image_, method='rosin', invert=True):
     return spots
 
 
+def get_well_mask(image_, segmethod='rosin'):
+
+    well_mask = ~thresh_and_binarize(image_, method=segmethod, invert=True)
```
bryantChhun:
```
~thresh_and_binarize(image_, method=segmethod, invert=True)
```
is equivalent to
```
thresh_and_binarize(image_, method=segmethod, invert=False)
```

Here I would prefer to use invert=False if only because it is already an argument to the function

run_array_analyzer.py
```diff
@@ -42,15 +50,16 @@ def parse_args():
     return parser.parse_args()
 
 
-def run_workflow(input_dir, output_dir, method, debug=False):
+def run_workflow(input_dir, output_dir, format, array_detection, debug=False):
```
bryantChhun: I think i prefer `method` over `array_detection`.  `method` could apply to any format array.  For example, we would calculate mean instead of median (or something else) for `well` format

bryantChhun (commented):
I've already tried merging with Jenny's #34 robustspotsdetection -- there are very few issues, only at cli parsing, but everything runs as normal.

Much of this could be integrated into one workflow in the future refactor.

array_analyzer/extract/image_parser.py
ieivanov: Fixed

array_analyzer/workflows/well_wf.py
ieivanov: Added new method `method == 'crop'` to `well_wf`. https://github.com/czbiohub/serology-COVID19/blob/aff4081e1e94d5d89de6c970230a739c03d3461d/array_analyzer/workflows/well_wf.py#L52-L61

Also added OD computation if there is a sample called 'Blank'.  The intensity of all blank wells is averaged. https://github.com/czbiohub/serology-COVID19/blob/aff4081e1e94d5d89de6c970230a739c03d3461d/array_analyzer/workflows/well_wf.py#L84-L91

run_array_analyzer.py
ieivanov: How about this? https://github.com/czbiohub/serology-COVID19/blob/aff4081e1e94d5d89de6c970230a739c03d3461d/run_array_analyzer.py#L64-L90

run_array_analyzer.py
ieivanov: How about calling it `workflow`? https://github.com/czbiohub/serology-COVID19/blob/aff4081e1e94d5d89de6c970230a739c03d3461d/run_array_analyzer.py#L64-L90

bryantChhun: > How about calling it `workflow`?

I think that's fine.  These flags can be adjusted later as more features are added.

bryantChhun (approved):
I think this is good to go.  It should be merged before robust spot detection.

bryantChhun:
I've tested this against both interp and fit April 4th flu data.  It ran without issue.  Will merge now

Issue #43, opened 2020-04-10
Add a check on the quality of spot detection and switch to the alternative approach if failed


smguo:
The failed wells in `fit` workflow seem to complement the failed wells in `interpolation` workflow. This is probably because `fit` approach tries to fit a full grid to the spots, but the fit can fail when only a few spots are present in the image. Interpolation approach on the other hand finds the x,y boundary of the grid and interpolates the grid positions. When fit approach converges to the right answer the determined spot positions seem more accurate than interpolation approach because fit approach fits to all the available spots but interpolation only uses the corner spots. 

I think we can have a check on the quality of spot detection and automatically switch to the alternative approach when detection using one approach fails. Some examples:

Fit approach:
![C3_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025023-784ebd80-7b39-11ea-8cb6-299e1c150e49.png)
![D3_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025049-91576e80-7b39-11ea-97e2-2ee4f8305983.png)
![D1_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025142-ca8fde80-7b39-11ea-8620-cc2e0dcd1af6.png)
![C11_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025164-d7143700-7b39-11ea-9c6a-faf9c9989126.png)

Interpolation approach:
![C3_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025236-02972180-7b3a-11ea-925a-36e778f75aa1.png)
![D3_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025256-0fb41080-7b3a-11ea-992f-2da3ff2c55a5.png)
![D1_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025271-1cd0ff80-7b3a-11ea-9596-a7bfe859dcfe.png)
![C11_overlayCentroids](https://user-images.githubusercontent.com/18110744/79025289-2bb7b200-7b3a-11ea-9f74-0ced2a2be3d6.png)

Fit approach output:
ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/Automated_Analysis_Spot-Fitting/4_10_13_21_19

interpolation approach output (using spot_intensity_fix branch):
[...]

PR #45, opened 2020-04-13
Spot intensity fix fit

jennyfolkesson:


array_analyzer/extract/img_processing.py
```diff
@@ -135,7 +139,9 @@ def get_spot_coords(im,
                     min_area=50,
                     max_area=10000,
                     min_circularity=0.1,
-                    min_convexity=0.5):
+                    min_convexity=0.5,
+                    minDistBetweenBlobs=10,
```
jennyfolkesson: Detail, but I (and PEP) prefer variables outside of libraries to all be snake case.

array_analyzer/extract/img_processing.py
```diff
@@ -163,11 +169,16 @@ def get_spot_coords(im,
...
+    params.minRepeatability = minRepeatability
+
 
     detector = cv.SimpleBlobDetector_create(params)
 
     # Normalize image
-    im_norm = ((im - im.min()) / (im.max() - im.min()) * 255).astype(np.uint8)
+    # im_norm = ((im - im.min()) / (im.max() - im.min()) * 255).astype(np.uint8)
+    # im_norm = im
+    im_norm = cv.GaussianBlur(im, (31, 31), 0)
```
jennyfolkesson: Maybe make sigma be an input parameter too.

array_analyzer/extract/img_processing.py
```diff
@@ -268,3 +279,77 @@ def crop_image_from_coords(im, grid_coords, margin=200):
...
+        thresh = threshold_minimum(image_, nbins=512)
+
+        spots = copy(image_)
+        spots[image_ < thresh] = 0
+        spots[image_ >= thresh] = 1
+
+    elif method == 'otsu':
+        spots = create_otsu_mask(image_, scale=1)
+
+    # elif method == 'multi_otsu':
```
jennyfolkesson: Will you need this commented out code? I'd say use it or lose it. :-)

array_analyzer/transform/array_generation.py
```diff
@@ -52,3 +55,116 @@ def build_centroid_binary_blocks(cent_list, image_, params_, return_type='region
...
+        background image without spots
+    :param params: dict
+        parameters parsed from the metadata
+    :return: dict
+        dictionary with format (row index, column index): prop
+    """
+
+    # values in mm
+    spot_width = params['spot_width']
+    PIX_SIZE = params['pixel_size_scienion']
```
jennyfolkesson: I'd reserve all caps for constants defined at the top of file.s

array_analyzer/extract/img_processing.py
```diff
@@ -268,3 +279,77 @@ def crop_image_from_coords(im, grid_coords, margin=200):
...
+
+def crop_image(arr, cx_, cy_, radius_, border_=200, last_pix=False):
+    """
+    crop the supplied image to include only the well and its spots
+
+    :param arr: image
+    :param float cx_: Center x coordinate
+    :param float cy_: Center y coordinate
+    :param float radius_: Crop radius
+    :param int border_: Margin on each side in pixels
```
jennyfolkesson: last_pix is not in docstring.

array_analyzer/extract/img_processing.py
```diff
@@ -268,3 +279,77 @@ def crop_image_from_coords(im, grid_coords, margin=200):
...
+    :return np.array crop: Cropped image
+    """
+    cx_ = int(np.rint(cx_))
+    cy_ = int(np.rint(cy_))
+    radius_ = int(radius_)
+    if last_pix:
```
jennyfolkesson: Not sure what last_pix is doing here? And don't you want to check to make sure that your bounding box is inside the image?
Why use a radius, do you always want a square cropped image?

array_analyzer/extract/img_processing.py
```diff
@@ -268,3 +279,77 @@ def crop_image_from_coords(im, grid_coords, margin=200):
...
+        bbox = [cy_ - (radius_ - border_),
+                cx_ - (radius_ - border_),
+                cy_ + (radius_ - border_) + 1,
+                cx_ + (radius_ - border_) + 1]
+    else:
+        bbox = [cy_ - (radius_ - border_),
+                cx_ - (radius_ - border_),
+                cy_ + (radius_ - border_),
+                cx_ + (radius_ - border_)]
+    crop = arr[bbox[0]:bbox[2], bbox[1]:bbox[3]]
```
jennyfolkesson: You're performing the radius - border operation four times, maybe just precompute that and skip the bbox altoghether?

array_analyzer/transform/point_registration.py
```diff
@@ -188,11 +191,15 @@ def particle_filter(fiducial_coords,
         # plt.show()
 
         min_dist = np.min(dists)
-        print(min_dist)
+        mean_dist = np.mean(dists)
```
jennyfolkesson: I might not have used the best convergence criterion, but I'm not sure we want to stop when the mean distance of the particle cloud isn't changing... I get that a stable point cloud is what will happen during convergence but we're optimizing for the minimum distance of the best particle and the stopping criterion should ideally reflect that.

array_analyzer/transform/array_generation.py
```diff
@@ -52,3 +55,116 @@ def build_centroid_binary_blocks(cent_list, image_, params_, return_type='region
...
+    for coord in coords:
+        cen_y, cen_x = coord
+        # convert the centroid position to an integer that maps to array indices
+        grid_y_idx = int(round((n_rows - 1) * ((cen_y - y_min) / y_range)))
+        grid_x_idx = int(round((n_cols - 1) * ((cen_x - x_min) / x_range)))
+        grid_id = (grid_y_idx, grid_x_idx)
+        label = 0
+        coord = [grid_id[0]/(n_rows - 1) * y_range + y_min,
+                         grid_id[1]/(n_cols - 1) * x_range + x_min]
+        # make bounding boxes larger to account for interpolation errors
```
bryantChhun, line 105: I believe lines 79 - 107 are repeated code from what is commented out in the main workflow.  Above line 107 you've created an an array index per centroid location and increased the size of the bounding box.

Line 200 of `registration_workflow` already has an array of RegionProps objects that you can iterate through.  I believe it would be much cleaner to use those.  If those RegionProp objects have bounding boxes that are too small, this is a simple parameter adjustment to registration_workflow(line 158) - array_gen.build_centroid_binary_blocks.

Essentially this function could reduce to:
1) receive the array from `props_array_placed` variable from `registration_workflow`line 200
2) for each RegionProps in 1, get its underlying intensity_image, then call threshold/generate_props as you do at and below line 145.  Replace the existing regionprops with the newone.

I could start a branch and submit PR to here if that'd be clearer

EDIT:
After some discussion with Syuan-Ming, I realize now that the "props_array" data is not a useful one.  It would be more intuitive to keep the dictionary structure, as Syuan-ming uses, and write props/ods directly from those.  Dictionary structure is also more translatable to pandas dataframes.

array_analyzer/transform/array_generation.py
```diff
@@ -52,3 +55,116 @@ def build_centroid_binary_blocks(cent_list, image_, params_, return_type='region
         return target*image_
     elif return_type == 'region':
         return target
+
+def coord_to_spot_int(coords, im_int, background, params):
```
bryantChhun: I think function names should start with "action" words or verbs.
To me, this function most specifically takes coordinates and returns regionprops of the coordinate's spot.
How about "generate_spot_regionprops"?

array_analyzer/extract/img_processing.py
smguo: I copied the parameter names from OpenCV blob detector. Will change it to snake case.

array_analyzer/extract/img_processing.py
smguo: Will clean up the commented unused code.

array_analyzer/extract/img_processing.py
smguo: The current implementation of `crop_image` will always output images with size that are even numbers. I added `last_pix` to add one more pixel at the right and bottom to make it odd numbers so the image can have the same size as the circular mask generated using disk structure element (which is always odd number)... I think I will rewrite this function.

array_analyzer/extract/img_processing.py
smguo: I was trying to reuse the existing function in the repo. I'm not using `border` at all. Will get rid of it.

array_analyzer/transform/point_registration.py
smguo: I made this change to be consistent with the change I made later on to output the mean particle rather than the best particle when I was troubleshooting failed spot detection. I was hoping using the mean particle (i.e. the mean of the posterior distribution) will be more robust than using the best particle (i.e. the point estimate at maximal posterior probability), but it turns out it doesn't make much difference in this case. I'll revert it back to best particle

jennyfolkesson:
@smguo thanks! Feel free to merge to robustspotdetection when you're ready.

bryantChhun (changes requested):
Overall the major contribution of this PR is to "refine" the current measurement.  Currently we use the image properties of the bounding box.  This PR introduces a segmentation step to extract the spot within the bounding box, then return that regionprops

array_analyzer/transform/array_generation.py
smguo: Yes I think pandas dataframe or dictionary would be a more suitable data structure than numpy arrays for this type of data. I can create an issue on this

array_analyzer/transform/array_generation.py
smguo: How about `get_spot_int()` or `get_spot_intensity()`? If we change the data structure to dataframe, the function would extract useful properties from regionprops objects, save them in the dataframe, and return the dataframe rather than passing regionprops objects

bryantChhun: Either of those is fine.  I think we should leave the data structure as is and adjust that in a later PR

smguo:
@jennyfolkesson and @bryantChhun I incorporated comments from both of you. Will merge it.

smguo:
Hey @jennyfolkesson I saw that you closed #43, but the feature I suggested hasn't been implemented yet. In my PR I only fixed the the spot detection issue for COVID_0408 data. I didn't test the fix on the older data and I'm not sure if the current fix won't fail again in the future dataset. 

My idea was to have a check on spot detection quality from fit approach (e.g. min distance output by particle filter), and run interpolation when it doesn't pass the check. Let me know what you think. This feature should be a separate PR.

jennyfolkesson:
Sorry! I reopened the issue. I just detected that spot detection missed a
fiducial in the March 25 plate so I'll look over the spot detection once
more.

Issue #29, opened 2020-04-06
ideas for refactoring: image_parser module is too large


bryantChhun:
The extract.image_parser module has become a catch-all of functions.  Many of these can be moved out.

I'll start by doing the following across two PRs:
1) copying functions to new modules in appropriate locations
2) changing import statements to the new modules/locations
3) finally, remove the old functions from the old modules (maybe a 3rd PR?)

mattersoflight:
I suggest creating one PR with refactor to which multiple individuals contribute and initiate one or two PRs. Our goal should be to describe dataflow with names of the modules/functions.  

For example,

Load images and metadata, write output:

io.load_scienion(path, metadata)
io.load_octopi(path, metadata)
io.save_OD()
io.save_intensities()
io.save_background()

Image->OD per well:

analyze.segment_well
analyze.compute_background
analyze.find_array_spots
analyze.compute_OD

debug plots:
debug.plot_overlay
debug.plot_icp

OD->calls across wells:
measure.standard_curve
measure.plot_OD(antigen,serum)
measure.antibody_concentration

The last module (measure ) needs to be defined.

bryantChhun:
> I suggest creating one PR with refactor to which multiple individuals contribute and initiate one or two PRs. Our goal should be to describe dataflow with names of the modules/functions.

One difficulty with this approach is that there can not be any permission, PR, or other check that someone pushes changes we all agree upon.  Contributors to this branch will simply "push to remote".  And all other contributors will have to frequently "pull from origin" and resolve conflicts to make sure we are synched with new pushes.

It is possible to do PR to this branch.  I think we each can have a repo fork and submit a PR across forks/branches.

PR #49, opened 2020-04-21
Robustspotdetection for master

bryantChhun:
Merge master into robustspotdetection.  The main adjustments are:
- introduction of well_wf.py workflow
- adjustment of the CLI flags.  Now uses "--workflow" or "-wf", instead of "--method" or "-m"
- bug fix in well_wf.py for "well_crop"
- some new functions in image_parser.py that aid well_wf.py

### DIFF
--- a/array_analyzer/extract/image_parser.py
+++ b/array_analyzer/extract/image_parser.py
@@ -117,6 +117,29 @@ def thresh_and_binarize(image_, method='rosin', invert=True):
     return spots
 
 
+def get_well_mask(image_, segmethod='rosin'):
+
+    well_mask = ~thresh_and_binarize(image_, method=segmethod, invert=True)
+
+    # Now remove small objects.
+    str_elem_size = 3
+    str_elem = disk(str_elem_size)
+    well_mask = binary_opening(well_mask, str_elem)
+
+    labels = measure.label(well_mask)
+    props = measure.regionprops(labels)
+
+    props = select_props(props, attribute="area", condition="greater_than", condition_value=10 ** 5)
+    well_mask[labels != props[0].label] = 0
+
+    return well_mask
+
+def get_well_intensity(image_, mask_):
+
+    well_int = np.median(image_[mask_])
+
+    return well_int
+
 def find_well_border(image, segmethod='bimodal', detmethod='region'):
     """
     finds the border of the well to motivate future cropping around spots
--- a/array_analyzer/extract/image_parser.py
+++ b/array_analyzer/extract/image_parser.py
@@ -64,52 +64,6 @@ def read_gray_im(im_path):
     return im
 
 
-# def thresh_and_binarize(image_, method='rosin', invert=True):
-#     """
-#     receives greyscale np.ndarray image
-#         inverts the intensities
-#         thresholds on the minimum peak
-#         converts the image into binary about that threshold
-#
-#     :param image_: np.ndarray
-#     :param method: str
-#         'bimodal' or 'unimodal'
-#     :return: spots threshold_min on this image
-#     """
-#
-#     if invert:
-#         image_ = u.invert(image_)
-#
-#     if method == 'bimodal':
-#         thresh = threshold_minimum(image_, nbins=512)
-#
-#         spots = copy(image_)
-#         spots[image_ < thresh] = 0
-#         spots[image_ >= thresh] = 1
-#
-#     elif method == 'otsu':
-#         spots = create_otsu_mask(image_, scale=1)
-#
-#     # elif method == 'multi_otsu':
-#     #     n_class = 3
-#     #     spots = create_multiotsu_mask(image_, n_class=n_class, fg_class=n_class - 1)
-#
-#     elif method == 'rosin':
-#         spots = create_unimodal_mask(image_, str_elem_size=3)
-#
-#     elif method == 'bright_spots':
-#         spots = image_ > np.percentile(image_, 95)
-#         str_elem = disk(10)
-#         # spots = binary_closing(spots, str_elem)
-#         spots = binary_opening(spots, str_elem)
-#         spots = clear_border(spots)
-#
-#     else:
-#         raise ModuleNotFoundError("not a supported method for thresh_and_binarize")
-#
-#     return spots
-
-
 def get_well_mask(image_, segmethod='rosin'):
 
     well_mask = thresh_and_binarize(image_, method=segmethod, invert=False)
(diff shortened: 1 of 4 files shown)


array_analyzer/extract/image_parser.py
```diff
@@ -64,6 +64,31 @@ def read_gray_im(im_path):
     return im
 
 
+def get_well_mask(image_, segmethod='rosin'):
+
+    well_mask = thresh_and_binarize(image_, method=segmethod, invert=False)
+
+    # Now remove small objects.
+    str_elem_size = 3
```
jennyfolkesson: How about moving this to be an input? And maybe add a docstring.

array_analyzer/extract/image_parser.py
```diff
@@ -115,6 +140,27 @@ def find_well_border(image, segmethod='bimodal', detmethod='region'):
     return [cy, cx], radii, well_mask
 
 
+def crop_image(arr, cx_, cy_, radius_, border_=200):
```
jennyfolkesson: What exactly goes into image_parser vs img_processing?
If image parser contains more spot/well specific function maybe cropping goes in img_processing?

array_analyzer/extract/image_parser.py
```diff
@@ -64,6 +64,31 @@ def read_gray_im(im_path):
...
+    props = measure.regionprops(labels)
+
+    props = select_props(props, attribute="area", condition="greater_than", condition_value=10 ** 5)
+    well_mask[labels != props[0].label] = 0
+
+    return well_mask
+
+
+def get_well_intensity(image_, mask_):
+
```
jennyfolkesson, line 102: I know this function is pretty self explanatory, but a docstring never hurts...

array_analyzer/workflows/well_wf.py
```diff
@@ -0,0 +1,109 @@
...
+    # sort by letter, then by number (with '10' coming AFTER '9')
+    well_dirs.sort(key=lambda x: (x[0], int(x[1:-7])))
+
+    # well_dirs = ['A7-Site_0']
+
+    int_well = []
+    for well_dir in well_dirs:
+
+        # read image
+        well_image_dir = [file for file in os.listdir(os.path.join(input_folder_, well_dir))
```
jennyfolkesson: First it finds a number of directories in input_folder_ then searches each directory for images and selects the first image? well_image_dir makes it sound like it's a directory you're getting.

array_analyzer/workflows/well_wf.py
```diff
@@ -0,0 +1,109 @@
...
+                img_ = im_crop.copy()
+                img_[~well_mask] = 0
+            else:
+                raise NotImplementedError(f'method of type {method} not supported')
+            io.imsave(output_name + "_masked_image.png",
+                      (img_/256).astype('uint8'))
+
+
+
+    df_int = pd.DataFrame(np.reshape(int_well, (8, 12)), index=list(string.ascii_uppercase[:8]), columns=range(1,13))
```
jennyfolkesson, line 102: Would it make sense to initialize a dataframe instead of a list before looping and just fill the dataframe with the intensity values?

run_array_analyzer.py
```diff
@@ -25,11 +26,11 @@ def parse_args():
         help="Output directory path",
     )
     parser.add_argument(
-        '-m', '--method',
+        '-wf', '--workflow',
```
jennyfolkesson, line 29: Have you tested this? I thought flags could only have one letter, not two.

run_array_analyzer.py
```diff
@@ -25,11 +26,11 @@ def parse_args():
...
         type=str,
-        choices=['interp', 'fit'],
-        default='interp',
-        help="Method for detected spots: interpolation or registration",
+        choices=['well_segmentation', 'well_crop', 'array_interp', 'array_fit'],
+        default='array_interp',
+        help="Analysis workflow",
```
jennyfolkesson: Maybe elaborate a bit on the help?

run_array_analyzer.py
```diff
@@ -42,14 +43,15 @@ def parse_args():
     return parser.parse_args()
 
 
-def run_workflow(input_dir, output_dir, method, debug=False):
+def run_workflow(input_dir, output_dir, workflow, debug=False):
     """
     For each image in input directory, run either interpolation (default)
     or registration of fiducials workflow.
```
jennyfolkesson, line 51: Describe the two new workflows.

run_array_analyzer.py
```diff
@@ -42,14 +43,15 @@ def parse_args():
...
     An xlsx file (and potentially debug plots) will be written to output directory.
 
     :param str input_dir: Input directory path
     :param str output_dir: Output directory path
+    :param str format: array or well
     :param str method: interp (interpolation) or fit (registration of fiducials)
```
jennyfolkesson: Change to workflow

array_analyzer/workflows/well_wf.py
```diff
@@ -0,0 +1,109 @@
...
+from datetime import datetime
+import skimage.io as io
+import pandas as pd
+import string
+import re
+import numpy as np
+
+
+def well_analysis(input_folder_, output_folder_, method='segmentation', debug=False):
+
```
jennyfolkesson, line 28: This function needs a docstring to explain what's happening in this workflow.

jennyfolkesson (changes requested):
Thanks incorporating changes from master! I have a few comments, let me know when you want me to look at it again.

array_analyzer/extract/image_parser.py
bryantChhun: I actually wanted to move this to img_processing.  Only reason I didn't was simply to keep the delta low for this PR.  Let's move it to img_processing for this PR.

But to answer your question, image_parser.py was meant to be just for image input/output.  It became a bit of a catch-all over time, and needs to be refactored further.  We should discuss in issue #29

array_analyzer/workflows/well_wf.py
bryantChhun: I agree it makes more sense to rename this to file.

Some background: this file structure (Leica data) is different from data from the Scienion.  Each image is in its own directory named "Site_0", "Site_1", etc..  In the future, we'll need to include this for all workflows.

run_array_analyzer.py
bryantChhun: yup.  It works with this argparse.ArgumentParser() It may not work with the older style "getopt" arg parser.

array_analyzer/workflows/well_wf.py
bryantChhun: I don't know pandas well enough to speak to this.  I assume the reshape helps us easily format the intensities into row-column format?  Otherwise the loops will have to know when to append a new row to the dataframe?  Not sure how this is done in pandas.

jennyfolkesson (approved):

PR #51, opened 2020-04-24
give cli a bit more intuitive name.

mattersoflight:


tests/run_array_analyzer_tests.py
```diff
@@ -3,7 +3,7 @@
 import unittest
 from unittest.mock import patch
 
-import run_array_analyzer as array_analyzer
+import pysero_analyzeOD as array_analyzer
```
jennyfolkesson: Could we make it snake case please? E.g. pysero_analyze_od?

smguo: How about `extract_od.py` or `run_od_extraction.py`? `analyzeOD` sounds to me more like interpret OD values rather than compute OD values. Also not sure if it's necessary to repeat the repo name `pysero`

mattersoflight: We want the CLI to be pip installable in the end like @bryantChhun figured out for another repo. So having `pysero` in the name is a good idea. How about `pysero_extract_od` for now? 

In the end both CLIs (current one and the one @smguo plans to work on)  should be callable as pysero  and each phase of pipeline can become a required argument. e.g. 
`pysero --extract_od` and `pysero --analyze_binding`. 

Thoughts?

mattersoflight: I have added two mutually exclusive arguments: -e runs the workflow as implemented now and -a is a placeholder for analyze_od, which indeed sounds more like interpretation.

jennyfolkesson (approved):

jennyfolkesson (approved):
Looks good. Could you please format the new arguments like the rest of the arguments in argparse, and add a help line before you merge? Thanks!

tests/run_array_analyzer_tests.py
smguo: I would suggest using subparser instead since `extract_od` and `analyze` will have different sets of arguments and should be separate:
https://stackoverflow.com/questions/17073688/how-to-use-argparse-subparsers-correctly
but this implementation works too.

smguo (approved):

bryantChhun (approved):
I have no opinion regarding using 
```
parser.add_mutually_exclusive_group
```
vs
```
parser.add_subparsers  # as per Syuan-Ming's suggestion
```

I might lean towards subparsers if only because it enables us to run both pipelines (analysis and interpretation) in one command.

mattersoflight:
Good idea to use subparser, Syuan-Ming. I will edit tomorrow.

sent on the move, please excuse typos

mattersoflight:
Thanks for the ideas. I tried the subparser, but didn't get the right behavior (See issue #55). Merging this PR now. @jennyfolkesson  please merge the PR on unittests next.

PR #54, opened 2020-04-27
Find images

jennyfolkesson:
Added support to read either a directory containing images or a number of subdirectories containing one image each.

I also added tests to the code, which I recommend doing moving forward for those who are so inclined. It really helps if the repo is intended to have many collaborators.


mattersoflight:
@jennyfolkesson I tried to run the pipeline with data acquired on Leica. I get following error in background estimation. 

image

mattersoflight:
I get an error with the other micro-manager dataset too (vanilla ELISA)
image

jennyfolkesson:
@mattersoflight this branch has not been tested on Leica images, and probably can't until metadata is refactored.
The first error is probably due to Leica images are 512 x 512, and then they are cropped to who knows what and after that background estimation is done with a block size of 128, which is probably larger than the image at that point.
So we probably need to add background block size as one of the parameters that need to be adjusted to physical size instead of pixel size.

The second error is because I only looked at interpolation and registration workflows...sorry about that. I'll update well workflow right away!

mattersoflight:
@jennyfolkesson I see that the error in well workflow came because of refactoring of the image io methods.

It is great that you started to refactor and are now adding tests in pytest format. 

Please implement the  following to make sure changes from this branch are useful in master:
* rebase onto current master and make sure that all tests use pytest library.
*  test array workflows in this branch with Scienion format `/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images` 
* test well workflows in this branch with `/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_leica/20200408_VanillaELISA_Leica2.5x_2` 
* If you like, curate the 5 'best images' and the 5 'worst images' for extract_od  workflows (`/Volumes/GoogleDrive/My Drive/ELISAarrayReader/data_for_tests_and_github/subset_for_tests`).

jennyfolkesson:
@mattersoflight I've rebased wrt master and checked that it works for array_fit and array_interp.
I checked that it ran for the well_* workflows but not sure what they're trying to do so I can't say they're working, just that they didn't crash. :-)
I also changed existing unittests to pytests.
I didn't pick images, figured that can be saved for another day/branch...
I also added code coverage, so we can see how testing will be getting better over time hopefully.

Please take a look again. Thanks!

array_analyzer/utils/io_utils.py
```diff
@@ -0,0 +1,97 @@
...
+        with one image each
+    :return dict well_images: Well name key, path to found image value
+    """
+    extensions = ('*.png', '*.tif', '*.jpg')
+
+    image_names = []
+    for ext in extensions:
+        search_str = os.path.join(input_dir, '**', ext)
+        image_names.extend(glob.glob(search_str, recursive=True))
+
```
mattersoflight: searching the folder tree recursively returns too many images, and assigns wrong images to the wells.
image

mattersoflight: Above screenshot was taken with following arguments: 
`-e
-i="/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images"
-o="~/Documents/images_local/2020-04-08-14-48-49-COVID_concentrationtest_April8_images"
-wf=array_fit`

mattersoflight (changes requested):
The discovery of image files is little buggy

mattersoflight:
I've fixed the bug by restricting the filename search to either the input folder or one level down from input folder. The well workflow runs fine now and array_fit workflow runs too, but I don't see an output OD file.

@jennyfolkesson  please take a look.

array_analyzer/utils/io_utils.py
jennyfolkesson: When I tested it on the data paths I didn't get these issues, but I never write outputs in the same directory as the input so that's probably why I missed that. Thanks for not only catching it but also fixing it!

jennyfolkesson:
@mattersoflight I did get an OD file, but I named it similar to the debug excel files. I've now renamed the OD excel file to match the one in the interpolation workflow: 'python_median_ODs.xlsx'. I hope that helps.

array_analyzer/utils/io_utils.py
bryantChhun: > When I tested it on the data paths I didn't get these issues, but I never write outputs in the same directory as the input so that's probably why I missed that. Thanks for not only catching it but also fixing it!

Somewhat related -- do we want to generally allow writing reports and outputs to the same directory as the raw data files?  It does cause problems when, for example, searching for excel metadata spreadsheets.

mattersoflight (approved):

mattersoflight:
I changed the other two filenames to be consistent too. Merging.

PR #57, opened 2020-04-30
Metadata encapsulation

bryantChhun:
## Purpose:
This PR introduces modules and functions to parse standardized metadata files.
** the standardized metadata file can be found here **
**/Volumes/GoogleDrive/My Drive/ELISAarrayReader/data_for_tests_and_github/Metadata_and_Plate_configuration.xlsx**

## Features:
This PR enables one to select the type of metadata format using the -m flag at CLI: 
- ".xml" files (old style) 
- ".xlsx" (excel)

It introduces two new modules:

> **extract/metadata.py**
> - contains a single class that will parse the correct metadata and assign constants

> **extract/constants.py**
> - holds hardware, array, and other constants in a single location.

## Details:
All workflows follow a similar procedure, which are now encapsulated in the MetaData class:
1) parse a text file containing hardware, array, experiment, data
2) place relevant info from 1 into dictionaries or lists
3) create arrays whose (row, column) positions contain information from 2
4) compute other constants (such as indexed fiducial locations) from 3

## Minor details:
- I've moved all notebooks to subfolders "notebooks".  There is one subfolder for each of extract/transform/load
- CI failed while reading excel files using Pandas.  It required the dependency: "xlrd >= 1.0.0"
- Writing temporary excel metadata files for tests, using Pandas, requires one of two backends: xlsxwriter, or openpyxl.  Thus, we should consider keeping openpyxl as a dependency (for tests, at least).
- "Well" segmentation for standard ELISAS uses a very simple workflow and mostly does not require any metadata from the above metadata sheet.  It is mostly untouched.

## Questions:
- A few experiments have a lot of .xlsx reports and extra files in the data folder.  Can we move these to subfolders and maintain a cleaner "data" directory?
- There are a few intermediate data structures (in constants.py):
``` layout, fiducials, spots, replicates ```
These are currently in constants.py but could easily be MetaData class variables, as they are not accessed anywhere else.  It's an easy refactor, but let me know if you think they'll be needed anywhere else.

## Future:
- there are very few tests at the moment.  I spent some time getting the conftest fixtures created, which should be a good starting point for making more tests.
- OD reporting is not formatted the per-antigen right now (it is per-well).  This will be fixed soon.

### CODE REVIEW

File: array_analyzer/extract/constants.py [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+FIDUCIALS_IDX = []
+
+# values used by point_registration.py
+SPOT_DIST_PIX = int()
+SPOT_DIST_UM = int()
+STDS = [500, 500, .1, .001]  # x, y, angle, scale
+REG_DIST_THRESH = 1000
+
+# constants for saving
+RUN_PATH = ''
```
jennyfolkesson, line 63: 💯

array_analyzer/extract/metadata.py
```diff
@@ -0,0 +1,221 @@
...
+        if repl_:
+            c.replicates = repl_
+        if params_:
+            c.params['rows'] = int(params_['rows'])
+            c.params['columns'] = int(params_['columns'])
+            c.params['v_pitch'] = float(params_['v_pitch'])
+            c.params['h_pitch'] = float(params_['h_pitch'])
+            c.params['spot_width'] = float(params_['spot_width'])
+            if c.METADATA_EXTENSION == 'xml':
+                c.params['pixel_size'] = c.params['pixel_size_scienion']
```
jennyfolkesson: Just for my understanding: is this because scienion is the only acquisition with xml metadata so that pixel size is known?

array_analyzer/extract/metadata.py
```diff
@@ -0,0 +1,221 @@
...
+        Calculate fiducial index like
+            FIDUCIALS_IDX = [0, 5, 6, 30, 35]\
+
+
+            FIDUCIALS_IDX = [0, 7, 8, 40, 47] for 8 columns
+        :return:
+        """
+        c.FIDUCIALS_IDX = list(np.where(c.FIDUCIAL_ARRAY.flatten() == 'Reference, Diagnostic')[0])
+        if len(c.FIDUCIALS_IDX) == 0:
+            c.FIDUCIALS_IDX = list(np.where(c.FIDUCIAL_ARRAY.flatten() == 'Fiducial')[0])
```
jennyfolkesson, line 189: 💯

array_analyzer/extract/metadata.py
```diff
@@ -0,0 +1,221 @@
...
+        # assuming similar v_pitch and h_pitch, average to get the SPOT_DIST
+        v_pitch_pix = v_pitch_mm/pix_size
+        h_pitch_pix = h_pitch_mm/pix_size
+        c.SPOT_DIST_PIX = np.mean([v_pitch_pix, h_pitch_pix]).astype('uint8')
+
+        # convert the SPOT_DIST to microns, 0 - 255
+        c.SPOT_DIST_UM = np.mean([v_pitch_mm * 1000, h_pitch_mm * 1000]).astype('uint8')
+
+    # set filesaving run_path
+    def _set_run_path(self, output_folder):
```
jennyfolkesson, line 209: Just FIY I added a similar function in find_images, but we can just remove whichever you prefer when merging branches.

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+        array_params['spot_width'] = float(layout['@expected_diameter'])
+        array_params['bg_offset'] = float(layout['@background_offset'])
+        array_params['bg_thickness'] = float(layout['@background_thickness'])
+        array_params['max_diam'] = float(layout['@max_diameter'])
+        array_params['min_diam'] = float(layout['@min_diameter'])
 
-    # replicates
-    repl = doc['configuration']['well_configurations']['configuration']['array']['spots']['multiplet']
+        # not present in .xml but manually added here:
+        array_params['pixel_size_scienion'] = 0.0049
```
jennyfolkesson: Should these values be located here or would it make sense to move them to constants?

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+    :param path_: list
+        of strings to the .csv paths
+    :return:
+    """
+    # assign names to each of the types == params, spot types, antigens
+    fiduc = list()
+    csv_antigens = list()
+    array_params = dict()
+
+    for meta_csv in path_:
```
jennyfolkesson, line 102: I believe you've previously checked that your dir contains the three separate csv files so you could just specifically read each of them instead of using elif statements...

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+
+                # parse antigens
+                elif "array_format_antigen" in meta_csv:
+                    for col, value in enumerate(row[1:]):
+                        pos = {'@row': str(row[0]),
+                               '@col': str(col - 1),
+                               '@antigen': str(value)}
+                        csv_antigens.append(pos)
+
+    return fiduc, None, csv_antigens, array_params
```
jennyfolkesson, line 130: Why are you returning a none?

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+
+
+def create_xlsx_dict(path_):
+    """
+    extracts fiducial, antigen, and array parameter metadata from .xlsx sheets
+    then populates dictionaries or lists with appropriate information
+    :param path_: path to .xlsx file
+    :return list fiduc: Fiducials and control info
+    :return list spots: None.  spots IDs not needed for .xlsx
```
jennyfolkesson, line 141: I see, so spot IDs are not needed for xlsx or csv dicts... which dicts are they needed for?

array_analyzer/workflows/interpolation_wf.py
```diff
@@ -1,49 +1,55 @@
...
+    # spot_ids = create_array(params['rows'], params['columns'])
+    # antigen_array = create_array(params['rows'], params['columns'])
 
     # adding .xml info to these arrays
-    spot_ids = populate_array_id(spot_ids, spots)
+    # spot_ids = populate_array_id(spot_ids, spots)
     # spot_ids = populate_array_fiduc(spot_ids, fiduc)
 
-    antigen_array = populate_array_antigen(antigen_array, spot_ids, repl)
+    # antigen_array = populate_array_antigen(antigen_array, spot_ids, repl)
```
jennyfolkesson: Do you intend to bring back the commented code in this PR or is it safe to delete?

array_analyzer/workflows/interpolation_wf.py
```diff
@@ -56,37 +62,30 @@ def interp(input_folder_, output_folder_, method='interp', debug=False):
...
+        spot_mask = img_processing.thresh_and_binarize(im_crop, method='bright_spots')
+        background = img_processing.get_background(im_crop, fit_order=2)
 
-        spot_props = generate_props(spot_mask, intensity_image_=im_crop)
+        spot_props = image_parser.generate_props(spot_mask, intensity_image_=im_crop)
 
         if debug:
-            well_path = os.path.join(run_path)
-            os.makedirs(run_path, exist_ok=True)
+            well_path = os.path.join(c.RUN_PATH)
```
jennyfolkesson: It looks like well_path and c.RUN_PATH is the same here.

array_analyzer/workflows/interpolation_wf.py
```diff
@@ -56,37 +62,30 @@ def interp(input_folder_, output_folder_, method='interp', debug=False):
...
+            os.makedirs(c.RUN_PATH, exist_ok=True)
```
jennyfolkesson: This dir has already been created before the loop.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -10,18 +10,20 @@
...
 import array_analyzer.transform.array_generation as array_gen
 
-FIDUCIALS = [(0, 0), (0, 1), (0, 5), (7, 0), (7, 5)]
-FIDUCIALS_IDX = [0, 5, 6, 30, 35]
-FIDUCIALS_IDX_8COLS = [0, 7, 8, 40, 47]
-SCENION_SPOT_DIST = 82
+# FIDUCIALS = [(0, 0), (0, 1), (0, 5), (7, 0), (7, 5)]
+# FIDUCIALS_IDX = [0, 5, 6, 30, 35]
+# FIDUCIALS_IDX_8COLS = [0, 7, 8, 40, 47]
+# SCENION_SPOT_DIST = 82
```
jennyfolkesson: I think these are safe to be removed thanks to your changes.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -114,13 +90,13 @@ def point_registration(input_folder, output_folder, debug=False):
             mean_point=mean_point,
             nbr_grid_rows=nbr_grid_rows,
             nbr_grid_cols=nbr_grid_cols,
-            spot_dist=SCENION_SPOT_DIST,
+            spot_dist=c.SPOT_DIST_PIX,
```
jennyfolkesson, line 75: 👍

tests/test_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+    with open(os.path.join(output_dir, 'second_xml.xml'), 'w') as fp:
+        pass
+    with pytest.raises(IOError):
+        MetaData(output_dir, output_dir)
+
+
+def test_xlsx(create_good_xlsx):
+    c.METADATA_EXTENSION = 'xlsx'
+    output_dir = create_good_xlsx
+    MetaData(output_dir, output_dir)
```
jennyfolkesson, line 61: 🥇

jennyfolkesson (changes requested):
This refactor looks great, it's so nice to get rid of those hardcoded constants (of which I'm guilty of adding). I just have some minor questions and comments.

mattersoflight:
Registration of spots looks good (from debug plots). 

But, there are differences in OD between when the pipeline was validated against scienion data and now.  Not sure if they are material.

Output that was compared against scienion (/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/python_fit_ODs.xlsx)
image

Output from current branch:
image


It is time to export ODs such that they are laid out per antigen with 2D format that represents the well. I am adding that information to metadata file. 

Some areas to pay attention to when rebasing this branch on current master:
* all workflows now use io_utils to find files either from folder or in subfolders.
* all workflows should output intensities by consistent name, so that interpretation part of the pipeline can read it all.
* if  rebasing the code breaks unit tests, please comment them out for now. They can be fixed in a dedicated branch once the squid data is analyzed.

mattersoflight:
Additional points:

* Keep the names of antigens precise (i.e., xkappa-biotin instead of Fiducial and GFP-foldon instead of negative) in `array_antigens` sheet. We can interpret antigens differently when analyzing the data. 

* Also, we should maintain index information for antigens when writing ODs, intensities, and background. 

See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.

array_analyzer/extract/metadata.py
bryantChhun: yes that is exactly right.  As far as I know, the sciReader doesn't even have other magnifications.  We've always seen the same pixel size

bryantChhun:
> Additional points:
> 
> * Keep the names of antigens precise (i.e., xkappa-biotin instead of Fiducial and GFP-foldon instead of negative) in `array_antigens` sheet. We can interpret antigens differently when analyzing the data.
> * Also, we should maintain index information for antigens when writing ODs, intensities, and background.
> 
> See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.

Would it be ok to allow both "Fiducial" and "xkappa-biotin" ?  I just wonder if a fiducial should always be made using xkappa-biotin.  I also wonder if "xkappa-biotin" is more prone to misspellings :-)

mattersoflight:
> > Additional points:
> > 
> > * Keep the names of antigens precise (i.e., xkappa-biotin instead of Fiducial and GFP-foldon instead of negative) in `array_antigens` sheet. We can interpret antigens differently when analyzing the data.
> > * Also, we should maintain index information for antigens when writing ODs, intensities, and background.
> > 
> > See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.
> 
> Would it be ok to allow both "Fiducial" and "xkappa-biotin" ? I just wonder if a fiducial should always be made using xkappa-biotin. I also wonder if "xkappa-biotin" is more prone to misspellings :-)

For this array, we could replace xkappa-biotin by fiducial. Not sure if there are other fiducial antigens. It is useful to leave the detail of antigen for whoever is interpreting the data.

bryantChhun:
> It is time to export ODs such that they are laid out per antigen with 2D format that represents the well. I am adding that information to metadata file.

Can we assume the image files are named a consistent way?  Such as "A1" "A2" ... "B12" ... "H12"?  I see there's some discussion about this with octopi, but don't know if there's a hard conclusion

smguo:
> > Additional points:
> > 
> > * Keep the names of antigens precise (i.e., xkappa-biotin instead of Fiducial and GFP-foldon instead of negative) in `array_antigens` sheet. We can interpret antigens differently when analyzing the data.
> > * Also, we should maintain index information for antigens when writing ODs, intensities, and background.
> > 
> > See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.
> 
> Would it be ok to allow both "Fiducial" and "xkappa-biotin" ? I just wonder if a fiducial should always be made using xkappa-biotin. I also wonder if "xkappa-biotin" is more prone to misspellings :-)

Maybe write both spot type and antigen name as 2 separate sheets in the output xlsx? eg. antigen name : xkappa-biotin, spot type: Fiducial

mattersoflight:
> Can we assume the image files are named a consistent way? Such as "A1" "A2" ... "B12" ... "H12"? I see there's some discussion about this with octopi, but don't know if there's a hard conclusion

Yes, we can safely assume that images will follow names of the well. It is a simple change/script for octopi to rename files A1...

bryantChhun:
> Maybe write both spot type and antigen name as 2 separate sheets in the output xlsx? eg. antigen name : xkappa-biotin, spot type: Fiducial

I'm ok with this.  This is actually the way scienion sciReader writes it.
It has a "marker" field in the .xml for "Reference, Diagnostic" = Fiducial, and "Diagnostic" for antigens.
Then later, it has a "multiplet" field for antigen name.

mattersoflight:
> > Maybe write both spot type and antigen name as 2 separate sheets in the output xlsx? eg. antigen name : xkappa-biotin, spot type: Fiducial
> 
> I'm ok with this. This is actually the way scienion sciReader writes it.
> It has a "marker" field in the .xml for "Reference, Diagnostic" = Fiducial, and "Diagnostic" for antigens.
> Then later, it has a "multiplet" field for antigen name.

updated the template.

bryantChhun:
> See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.

Right now we write three separate workbooks: ODs, INT, BG, with the latter two written only during "debug" mode.

I see that your template has three worksheets, one for: OD, INT, BG.  Perhaps it's better to separate the INT and BG into separate workbooks, like earlier?

mattersoflight:
> > See the output template in '/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-04-08-14-48-49-COVID_concentrationtest_April8_images/pysero_output_data_metadata.xlsx'. The first few sheets of workbook are metadata and the rest, od, intensity, background per antigen.
> 
> Right now we write three separate workbooks: ODs, INT, BG, with the latter two written only during "debug" mode.
> 
> I see that your template has three worksheets, one for: OD, INT, BG. Perhaps it's better to separate the INT and BG into separate workbooks, like earlier?

I suggest always exporting intensity and background - they are useful for interpretation.

array_analyzer/extract/constants.py
smguo: I like the idea of using module variables to share variables between different modules. Much cleaner!

mattersoflight:
"A few experiments have a lot of .xlsx reports and extra files in the data folder. Can we move these to subfolders and maintain a cleaner "data" directory?" 

I agree, we should organize the data directory. Please feel free to move files around on google drive.

array_analyzer/extract/metadata.py
```diff
@@ -0,0 +1,221 @@
...
+            fiduc_, spots_, repl_, params_ = txt_parser.create_xml_dict(xml_path)
+
+        elif c.METADATA_EXTENSION == 'well':
+            self._set_run_path(output_folder_)
+            return
+
+        elif c.METADATA_EXTENSION == 'csv':
+            # check that three .csvs exist
+            three_csvs = ['array_format_antigen', 'array_format_type', 'array_parameters']
+            csvs = [f for f in os.listdir(input_folder_) if '.csv' in f]
```
smguo, line 44: What is the csv metadata format for?

array_analyzer/extract/constants.py
```diff
@@ -0,0 +1,54 @@
...
+    'pixel_size_scienion': 0.0049,
+    'pixel_size_octopi': 0.00185,
+    'pixel_size': None
+}
+
+layout = dict()
+fiducials = dict()
+spots = dict()
+replicates = dict()
+
```
smguo, line 34: These seems to me more like intermediate variables that get converted to array variables below later. Maybe leave them at where the conversion happens?

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+                elif "array_format_parameters" in meta_csv:
+                    array_params[row[0]] = row[1]
+
+                # parse fiducials
+                elif "array_format_type" in meta_csv:
+                    for col, value in enumerate(row[1:]):
+                        pos = {'@row': str(row[0]),
+                               '@col': str(col - 1),
+                               '@spot_type': str(value)}
+                        fiduc.append(pos)
```
smguo, line 120: Is the csv metadata format also in array like xlsx format? Here you convert the metadata into dictionary first then later convert dictionaries into numpy arrays. But if the info is already in array format in the csv. You can convert them into numpy arrays directly using Pandas function without doing a two-step conversion

array_analyzer/extract/txt_parser.py
```diff
@@ -38,34 +53,124 @@ def create_xml_dict(path_):
...
+                pos = {'@row': row,
+                       '@col': col,
+                       '@spot_type': str(value)}
+                fiduc.append(pos)
+            else:
+                pos = {'@row': row,
+                       '@col': col,
+                       '@antigen': str(value)}
+                xlsx_antigens.append(pos)
+
```
smguo, line 178: Same comment here as for csv. I'm not sure the point to convert the 2D pandas table to a dictionary... maybe I'm missing something here

array_analyzer/extract/txt_parser.py
```diff
@@ -184,3 +289,20 @@ def populate_array_antigen(arr, id_arr_, repl):
...
+        list of dictionaries whose keys define array coordinates and values
+    :return:
+    """
+    for antigen in csv_antigens_:
+        r = antigen['@row']
+        c = antigen['@col']
+        v = antigen['@antigen']
+        arr[r, c] = v
+
+    return arr
```
smguo, line 340: There are many similar `populate_array_...` methods in this module to covert dictionaries into 2D arrays. Maybe consider to consolidate them in the next refactor if not in this PR. I believe you can achieve the same thing in 2 lines with pandas:
https://stackoverflow.com/questions/20638006/convert-list-of-dictionaries-to-a-pandas-dataframe
https://stackoverflow.com/questions/9339184/python-pandas-dataframe-convert-1column-data-into-2d-data-format

smguo (approved):
Looks great! the idea of constant module is neat. I left some comments. Some of them require substantial changes and could be addressed in the next refactor.

array_analyzer/extract/metadata.py
bryantChhun: good question.  This is something i wrote before we settled on xlsx.  It is not enabled, and I haven't removed it just in case.  But probably it'll disappear soon.

array_analyzer/extract/constants.py
bryantChhun: yes that is exactly right.  I put it up as one of the questions above.  It can easily become a local MetaData class variable.

array_analyzer/extract/txt_parser.py
bryantChhun: The .xml data is not in array format.  From my a quick search, it appears pandas can not read .xmls into dataframes either.  (please ignore the .csv parser, this PR focuses on only .xml and .xlsx)

But the decision with this function (txt_parser.create_xlsx_dict) was to stay consistent with existing machinery.  I do think you're right -- if we simply parse the .xlsx sheets directly into an array, it simplifies that parsing.  There's more to say here but let's leave it for place (a new issue?)

mattersoflight:
1. The constants module is really well thought.  Well done.
2.  I have tested the current version and the pipeline runs well with fit workflow. The ODs output per plate layout look good. I think placing OD, intensities, and background in separate excel file makes sense.  
3. I committed a change so that files are not output in pysero_... directory. Let's copy the metadata in there too. This is to clean up file organization and keep all files needed for interpretation in the same folder.
4. The background estimation in the interp workflow is out of sync with registraton (/Volumes/GoogleDrive/My Drive/ELISAarrayReader/data_for_tests_and_github/April_8_COVID_ConcentrationTest_metadata-branch/pysero_20200501_0829_interp). Also, there are slight differences in when intensities and background are generated. I suggest refactoring the code such that two workflows differ only in how the data is analyzed.

 @bryantChhun. I suggest addressing point-3  and then analyzing the squid data with this branch.  It is good to fix 4 if you or @smguo find time to implement and test. I will now be able to test tomorrow. Otherwise, I am okay with merging before 4 is fixed and turning this point into an issue to be addressed in later PR along with other issues.

array_analyzer/extract/txt_parser.py
bryantChhun: I thought about this more and tried something.  It's the new function txt_parser.create_xlsx_array.  It doesn't work right now, but could probably be fixed later on.  Let's revisit after this PR.

array_analyzer/extract/metadata.py
bryantChhun: I think i'm going to keep the version in MetaData.  Though, I can go either way if you have a preference.

array_analyzer/extract/metadata.py
bryantChhun: sorry, it is not used.  That requires a separate .csv for every sheet.  I thought it was going to be our approach, but I think excel .xlsx workbooks are better, especially with pandas

array_analyzer/extract/constants.py
bryantChhun: Fixed in a new commit

mattersoflight:
Strong work @bryantChhun! Merging.

PR #60, opened 2020-05-08
Fishy registration

jennyfolkesson:
I have worked on registration to make it more robust to the cuttlefish data as well as Jo's plates.

I had to relax the constraints on both angle and scale for particle stds, and then increase the number of particles since there's now more degrees of freedom.

I also switched out the top hat filter in spot detection for a Laplacian of Gaussian filter, and added an extra check that enough spots have been detected. If not, min repeatability will be decreased.

There are multiple fails in 2020-05-02-13-49-02-COVID_May2_JVassay_FcIgGplate_images, e.g. D12 and F1. As much as I welcome all kinds of image artifacts and smears that will help make the algorithm more robust, I think at some point for particularly bad images we'll have to start flagging fails. 
Right now I issue a warning if the final particle filter fails, but we should instead at that point stop analyzing, or return NaNs or something. I'd love to hear your thoughts on this.

Oh and by the way, @bryantChhun, I'm sorry but I renamed c to constants, turns out I can't handle single letter variables outside single functions well (so hard to search for!)...


jennyfolkesson:
@bryantChhun @smguo @mattersoflight I've tuned the spot registration again to work better with the May 2 data.

Main changes is that 
- I adapted spot detection settings to the imaging params instead of pixel sizes, 
- tuned LoG filters,
- Added well detection and cropping prior to spot detection
- made spot detection a class,
- will stop analysis if registration fails (ODs will not be written in these cases)

With the May 2 data it looks get_spot_intensity is a bit all over the place, but I didn't have time to investigate.

Please take a look again. Thanks!

bryantChhun (approved):
Looks good.  Tested on March 25, April 4th, April 8th, May 1st data without errors

PR #61, opened 2020-05-09
Fine tune fishy registration 

smguo:
The LoG filter @jennyfolkesson added in #60 works great. This PR adds a check to reduce sigma of LoG filter when registration failed to enhance the contrast of dim fiducials, and run registration again. This fixes the failed registration in dataset 2020-05-02-13-49-02-COVID_May2_JVassay_FcIgGplate_images ('A6', 'A11', 'D12', 'E1', 'E6', 'F1', 'F6'). Haven't tested this fix on cuttlefish data but happy to test if @jennyfolkesson you can share the pycharm run config. 

Also save LoG filtered image when run in debug mode. Found it useful for tuning size of LoG filter.

### DIFF
--- a/array_analyzer/workflows/registration_workflow.py
+++ b/array_analyzer/workflows/registration_workflow.py
@@ -52,13 +52,15 @@ def point_registration(input_dir, output_dir):
     # loop over well images
     # ================
     well_images = io_utils.get_image_paths(input_dir)
-
+    well_failed = []
+    # well_names = ['A6', 'A11', 'D12', 'E1', 'E6', 'F1', 'F6']
+    # well_images = {well: well_images[well] for well in well_names}
     for well_name, im_path in well_images.items():
         start_time = time.time()
         image = io_utils.read_gray_im(im_path)
 
         nbr_expected_spots = nbr_grid_cols * nbr_grid_rows
-        spot_coords = img_processing.get_spot_coords(
+        spot_coords, im_filtered = img_processing.get_spot_coords(
             image,
             min_area=500,
             min_thresh=100,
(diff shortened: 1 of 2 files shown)


array_analyzer/workflows/registration_workflow.py
```diff
@@ -52,13 +52,15 @@ def point_registration(input_dir, output_dir):
     # loop over well images
     # ================
     well_images = io_utils.get_image_paths(input_dir)
-
+    well_failed = []
+    # well_names = ['A6', 'A11', 'D12', 'E1', 'E6', 'F1', 'F6']
```
jennyfolkesson, line 56: Please remove commented out code.

array_analyzer/workflows/registration_workflow.py
```diff
@@ -96,20 +98,43 @@ def point_registration(input_dir, output_dir):
...
+            spot_coords, im_filtered = img_processing.get_spot_coords(
+                image,
+                min_area=200,
+                min_thresh=100,
+                max_thresh=255,
+                min_circularity=0.1,
+                min_convexity=0.5,
+                min_dist_between_blobs=10,
+                min_repeatability=2,
+                sigma_gauss=10,
```
jennyfolkesson, line 113: So the spot detection consistently fails for small spots? In a future PR I could try combining different scales in spot detection, but this is good for now...

array_analyzer/workflows/registration_workflow.py
```diff
@@ -96,20 +98,43 @@ def point_registration(input_dir, output_dir):
...
                 particles=particles,
                 stds=np.array(constants.STDS),
                 stop_criteria=0.1,
-                remove_outlier=True,
                 debug=constants.DEBUG,
             )
-            # Warn if fit is still bad
             if min_dist > constants.REG_DIST_THRESH:
-                warnings.warn("Final registration failed, registration may be flawed")
+                warnings.warn("Registration failed, repeat with outlier removal")
```
jennyfolkesson, line 125: It seems like a lot to repeat the particle filter 3 times with different conditions... I'd rather get the spot detection right in the first pass. Perhaps we can leave it in for now with a TODO to improve spot detection?

array_analyzer/workflows/registration_workflow.py
```diff
@@ -96,20 +98,43 @@ def point_registration(input_dir, output_dir):
...
+                    spot_coords=spot_coords,
+                    particles=particles,
+                    stds=np.array(constants.STDS),
+                    stop_criteria=0.1,
+                    remove_outlier=True,
+                    debug=constants.DEBUG,
+                )
+                if min_dist > constants.REG_DIST_THRESH:
+                    warnings.warn("Final registration failed, registration may be flawed")
+                    well_failed.append(well_name)
```
jennyfolkesson, line 137: Why collect well names in a list if all you'll do is print them? You might as well print them as they happen. Ideally we should set all OD values to None for failed wells so no one mistakenly analyzes them.

jennyfolkesson (approved):
Thanks for testing and improving it Syuan-Ming! I left some comments but overall it looks good.

array_analyzer/workflows/registration_workflow.py
smguo: Yes this is more a temporary fix because I didn't want to change the parameters you have tuned for spot detection to work for cuttlefish data, but this can be done inside spot detection

array_analyzer/workflows/registration_workflow.py
smguo: I agree. That would be ideal

array_analyzer/workflows/registration_workflow.py
smguo: Throw warnings when the registration fails doesn't seem to work every time for some reasons... when I ran it in pycharm sometimes the registration clearly failed but no warning was output in the console. I'm not sure why it happened but printing the warning at the end seems to work more robustly. Also the pipeline prints lots of output (min distance from particle filtering in every interation) when run in debug mode, which makes it hard to track which well has failed. I think it would be useful to write this info in a log file or the output metadata, in addition to setting the spot intensity to none...

array_analyzer/workflows/registration_workflow.py
jennyfolkesson: I did another iteration on the fishy registration PR to increase stability for the May 2 datasets. Sorry for making your changes out of sync! But I did find that smaller sigma worked better for most data thanks to your changes so thank you.

array_analyzer/workflows/registration_workflow.py
jennyfolkesson: I totally agree, I too have been thinking it would be great to add some logging. Hopefully it's something we can add once we have the more critical path things in place.

PR #66, opened 2020-05-19
when building a target image, must slice into it in [col, row] order

bryantChhun:
very small fix in the way the composite_spots_img is generated.

The target image needs to be sliced in [col, row] order.  This only influences the way debug data is generated.

### DIFF
--- a/array_analyzer/load/debug_plots.py
+++ b/array_analyzer/load/debug_plots.py
@@ -37,11 +37,12 @@ def assign_region(target_, props_, intensity_image_=None):
     :return:
     """
 
-    min_row, min_col, max_row, max_col = props_.bbox
+    # min_row, min_col, max_row, max_col = props_.bbox
+    min_col, min_row, max_col, max_row = props_.bbox
     if intensity_image_ is None:
-        target_[min_col:max_col, min_row:max_row] = props_.intensity_image
+        target_[min_row:max_row, min_col:max_col] = props_.intensity_image
     else:
-        target_[min_col:max_col, min_row:max_row] = intensity_image_[min_col:max_col, min_row:max_row]
+        target_[min_row:max_row, min_col:max_col] = intensity_image_[min_row:max_row, min_col:max_col]
     return target_
 
 
--- a/array_analyzer/load/debug_plots.py
+++ b/array_analyzer/load/debug_plots.py
@@ -39,10 +39,9 @@ def assign_region(target_, props_, intensity_image_=None):
 
     min_row, min_col, max_row, max_col = props_.bbox
     if intensity_image_ is None:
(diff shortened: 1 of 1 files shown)


array_analyzer/load/debug_plots.py
```diff
@@ -39,10 +39,9 @@ def assign_region(target_, props_, intensity_image_=None):
 
     min_row, min_col, max_row, max_col = props_.bbox
     if intensity_image_ is None:
-        target_[min_row:max_row, min_col:max_col] = props_.intensity_image
+        target_[min_col:max_col, min_row:max_row] = props_.intensity_image
```
jennyfolkesson: NumPy indexing is (row, column) (which we often call (y, x)) so you'll probably need to change the min max values you get from props_.bbox to reflect numpy indexing.

jennyfolkesson (changes requested):
Just wanted to make sure indexing is consistent...

array_analyzer/load/debug_plots.py
bryantChhun: Admittedly i'm getting turned about here.  If we switch the row/col from props_.bbox, the change to image slicing effectively goes back to the old broken way.  Do you mean we should swap
```
min_row, min_col, max_row, max_col = props_.bbox
```
to
```
min_col, min_row, max_col, max_row= props_.bbox
```
and keep the slicing like before?:
```
target_[min_row:max_row, min_col:max_col] = props_.intensity_image
```

jennyfolkesson: Yes, that's what I meant. Thanks!

jennyfolkesson (approved):

smguo:
Just want to make sure I understand this PR... Here is the [definition of regionprop bbox](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.regionprops) from scikit-image:
"""
bbox
Bounding box (min_row, min_col, max_row, max_col). Pixels belonging to the bounding box are in the half-open interval [min_row; max_row) and [min_col; max_col).
"""

So the current indexing seems fine to me as it is... unless there is a flip of the order of the elements in bbox somewhere after it was generated?

smguo (changes requested):
Can we double check the indexing of bbox as I commented above? Thanks!

bryantChhun:
Let’s hold off on merging this PR because Jenny made a lot of changes with indexing in the uint16 PR.  I think those fixed this problem but haven’t tested.

bryantChhun:
@smguo @jennyfolkesson 
I tested this problem on the new master and it is fixed.  Going to close this PR without merging.

If there's another discrepancy (say, master has somehow changed "find_spot_intensity" in a way that is not correct) let's make that a new PR.

PR #67, opened 2020-05-20
added integration config and array fit slow test

jennyfolkesson:
Adding CI for slow integration tests.

Added try catch to report since xlsx write fails.
Probably need to separate out the 6x8 sample for it to work...


bryantChhun (approved):
Looks good!  Will add appropriate annotations to new tests in the future.

PR #68, opened 2020-05-21
Uint16 fix

jennyfolkesson:
The latest cuttlefish data showed that there were some bugs since we've only run on uint8's, and some bugs due to indexing (which I didn't notice because I've been running a lot of 6 x 6 grids.

These are the kind of mistakes I make when I don't take the time to write tests while I write code. I think I'll make another PR with some tests added just to feel better. :-)

I've changed the code to get rid of y, x indexing in favor of rows, cols, which is more numpyesque.
And I've adapted spot detection to work with either uint8 or 16.
The get spot intensity function goes a bit crazy sometimes with the comets, but I think that's for another PR. Thanks!

### DIFF
--- a/array_analyzer/load/debug_plots.py
+++ b/array_analyzer/load/debug_plots.py
@@ -177,31 +177,34 @@ def plot_registration(image,
                       grid_coords,
                       reg_coords,
                       output_name,
+                      max_intensity=255,
                       margin=100):
     """
     Plots all detected spots, initial fiducial coordinates and registered grid.
 
     :param np.array image: Input image
-    :param np.array spot_coords: Detected spot coordinates (nbr spots x 2)
+    :param np.array spot_coords: Detected spot coordinates (nbr spots x rows, cols)
     :param np.array grid_coords: Initial estimate of fiducial coordinates
     :param np.array reg_coords: Registered coordinates
     :param str output_name: Path + well name, _registration.png will be added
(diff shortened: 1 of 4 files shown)


array_analyzer/load/debug_plots.py
```diff
@@ -177,31 +177,34 @@ def plot_registration(image,
...
     im_shape = image.shape
-    x_min = int(max(0, np.min(all_coords[:, 0]) - margin))
-    x_max = int(min(im_shape[1], np.max(all_coords[:, 0]) + margin))
-    y_min = int(max(0, np.min(all_coords[:, 1]) - margin))
-    y_max = int(min(im_shape[0], np.max(all_coords[:, 1]) + margin))
-    im_roi = image[y_min:y_max, x_min:x_max]
+    row_min = int(max(0, np.min(all_coords[:, 0]) - margin))
+    row_max = int(min(im_shape[0], np.max(all_coords[:, 0]) + margin))
+    col_min = int(max(0, np.min(all_coords[:, 1]) - margin))
+    col_max = int(min(im_shape[1], np.max(all_coords[:, 1]) + margin))
```
bryantChhun, line 200: I noticed that row-col im_shape indexing was flipped.  Glad to see it was not intentional!

bryantChhun (approved):
This is tested on April 4th, 8th, and Feb 3rd 6x8 data.  All ran smoothly.

PR #69, opened 2020-05-26
Spot intensity tune

jennyfolkesson:
Refactored get spot intensity to make search bounding box smaller so it doesn't "wander off" and misses the spots.


array_analyzer/transform/array_generation.py
```diff
@@ -86,80 +89,102 @@ def get_spot_intensity(coords, im_int, background, params):
...
+                mask=mask_spot,
+                bbox=bbox,
+                centroid=coord,
+            )
+            bg_prop.generate_props_from_disk(
+                image=bg_spot,
+                mask=mask_spot,
+                bbox=bbox,
+                centroid=coord,
+            )
```
bryantChhun, line 185: I'm not sure the logic is quite the same as what we've been using.  The values in the end may well be the same though.

The earlier version of get_spot_intensity would create a default disk size upon MockRegionProp init.  Then, when `thresh_and_binarize` then `generate_props` were called, if there is new location, the props.bbox would be resized to be smaller so it contains a uniform-intensity region only around the center of the spot.

This version is always producing a disk of a fixed size?

I don't think there's a "right" method here.  We are currently doing further testing to see how small changes like this influence the data.

EDIT: Ok Looking at the results, there's some amount of bounding box adjustment but not much anymore.

array_analyzer/transform/array_generation.py
```diff
@@ -86,80 +89,102 @@ def get_spot_intensity(coords, im_int, background, params):
...
+        )
+        # Create spot and background instance
+        spot_prop = regionprop.SpotRegionprop(
+            label=label,
+        )
+        bg_prop = regionprop.SpotRegionprop(
+            label=label,
+        )
+        # Mask spot should cover a certain percentage of ROI
+        if np.mean(mask_spot) > constants.SPOT_MIN_PERCENT_AREA:
```
bryantChhun, line 140: is ```np.mean(mask_spot)``` the average area, not the percent area?

array_analyzer/extract/img_processing.py
```diff
@@ -212,21 +213,22 @@ def crop_image_from_coords(im, grid_coords, margin=200):
...
     im_h, im_w = im.shape
     # truncate the bounding box when it exceeds image size
-    bbox = np.rint([max(cy - height / 2, 0),
-                   max(cx - width / 2, 0),
-                   min(cy + height / 2, im_h),
-                   min(cx + width / 2, im_w)]).astype(np.int32)
+    bbox = np.rint([max(c_row - height / 2, 0),
+                   max(c_col - width / 2, 0),
+                   min(c_row + height / 2, im_h),
+                   min(c_col + width / 2, im_w)]).astype(np.int32)
```
smguo, line 231: 👍

array_analyzer/transform/array_generation.py
bryantChhun: Ah!  mask_spot is just 0-1 so mean is percent.

array_analyzer/extract/img_processing.py
```diff
@@ -252,28 +254,47 @@ def crop_image(arr, cx_, cy_, radius_, border_=200):
...
+def get_largest_component(spot_segm):
+    """
+    Remove everything but the largest connected component from segmented image.
+
+    :param np.array spot_segm: Binary segmented 2D image
+    :return np.array largest_component: Largest connected component in image
+    """
+    labels = label(spot_segm)
+    largest_component = labels.copy()
+    if labels.max() > 0:
```
smguo, line 266: This is minor but I think you can do `if not spot_segm.any(): return spot_segm` to skip labeling the image if the image is all zeros

array_analyzer/extract/img_processing.py
```diff
@@ -284,13 +305,15 @@ def thresh_and_binarize(image_, method='rosin', invert=True, min_size=10, thr_pe
...
-        # spots = binary_closing(spots, str_elem)
         spots = binary_opening(spots, str_elem)
         spots = clear_border(spots)
 
     else:
         raise ModuleNotFoundError("not a supported method for thresh_and_binarize")
 
+    if get_lcc:
+        spots = get_largest_component(spots)
+
```
smguo, line 317: 👍

array_analyzer/transform/array_generation.py
```diff
@@ -86,80 +89,102 @@ def get_spot_intensity(coords, im_int, background, params):
...
+                image=bg_spot,
+                mask=mask_spot,
+                bbox=bbox,
+                centroid=coord,
+            )
+
+        spot_props[grid_id] = spot_prop
+        bg_props[grid_id] = bg_prop
+
+    return spot_props, bg_props
```
smguo, line 190: 💯

array_analyzer/utils/spot_regionprop.py
```diff
@@ -0,0 +1,86 @@
...
+        self.mean_intensity = skimage_props.at[0, 'mean_intensity']
+
+        self.centroid = [bbox[0] + skimage_props.at[0, 'centroid-0'],
+                         bbox[1] + skimage_props.at[0, 'centroid-1']]
+        self.bbox = [bbox[0] + skimage_props.at[0, 'bbox-0'],
+                     bbox[1] + skimage_props.at[0, 'bbox-1'],
+                     bbox[0] + skimage_props.at[0, 'bbox-2'],
+                     bbox[1] + skimage_props.at[0, 'bbox-3']
+                     ]
+        self.assign_median_masked()
```
smguo, line 87: I like the idea of getting rid of scikit-image regionprops objects entirely and redefine the property names... I found the property names of scikit-image regionprops unintuitive. This is better.

tests/utils/spot_regionprop_test.py
```diff
@@ -0,0 +1,88 @@
...
+    assert prop.label == 4
+    assert prop.image.all() == im_spot.all()
+    assert 29 <= prop.centroid[0] <= 30
+    assert 34 <= prop.centroid[1] <= 35
+    # There are the coordinates for the mask bounding box
+    assert prop.bbox == [18, 23, 42, 47]
+    assert prop.mask.all() == mask_spot.all()
+    assert prop.mean_intensity == np.mean(im_spot[mask_spot > 0])
+    assert prop.median_intensity == np.median(im_spot[mask_spot > 0])
+    assert prop.masked_image.all() == (im_spot * mask_spot).all()
```
smguo, line 88: 💯

smguo (approved):
Thanks for refactoring 'get_spot_intensity` @jennyfolkesson! It's much cleaner now. The PR looks great to me.

smguo:
Please feel free to merge after resolving the conflicts with master.

smguo (approved):
Please feel free to merge after resolving the conflicts with master.

bryantChhun (approved):
looks good.  Been running it together with nonetypes.  So far so good

array_analyzer/transform/array_generation.py
jennyfolkesson: It's producing a disk of fixed size if there's no mask detected in thresh and binarize. I'll merge as is for now but I'm open to reevaluating best strategies here.
For instance, the skimage clear_border might not be the best if the large ROI does not reach all the way around a spot that's not quite on the grid.

Issue #63, opened 2020-05-18
Write antigen report throws error


gt8mar:
**Describe the bug**
See screenshot below. The bug is caused because there is nothing in the antigen_position array. In the metadata, there are "empty" and "Blank" entries where there aren't antigens. Perhaps these should be " " or "None"? 

The function write_antigen_report() appears to have code to ignore the previous two entries, but not "empty" or "blank". 

**To Reproduce**
1. Run folder:
/ELISAarrayReader/images_cuttlefish/2020-05-02-13-49-02-COVID_May2_JVassay_FcIgGplate_images/rotated/
or 
"/ELISAarrayReader/images_cuttlefish/2020-05-01-17-29-54-COVID_May1_JBassay_images/rotated/
2. Run pysero.py using pysero-fishy_registration (5-15-2020)

**Expected behavior**
Error with screenshot below:

**Screenshots**
<img width="645" alt="Issue_error_05172020" src="https://user-images.githubusercontent.com/35577361/82165663-d2464f80-986a-11ea-9608-1d8ababe56d1.PNG">

**Operating environment (please complete the following information):**
 - OS: Windows, Master, branch=fishy_registration

**Additional context**
metadata file: "/ELISAarrayReader/images_cuttlefish/2020-05-02-13-49-02-COVID_May2_JVassay_FcIgGplate_images/rotated/pysero_output_data_metadata.xlsx"

bryantChhun:
I've isolated the offender -- it's well E5 in that data.  Specifically, the spot detection must be failing, resulting in "None" being written to the 96-well "od_well" array (it would also be "None" in Int-array and BG-array, if it got that far).

I think the solution here touches on other issues too -- we should report failed spots and handle them gracefully here.  There's another slight issue here too: the "od_well" array is an 8x12 array (96-wells) of `objects`.  It's initialized as such in metadata.py:

```
def _calc_empty_plate_const():

    constants.WELL_BG_ARRAY = np.empty((8, 12), dtype=object)
    constants.WELL_INT_ARRAY = np.empty((8, 12), dtype=object)
    constants.WELL_OD_ARRAY = np.empty((8, 12), dtype=object)
```

instead we could use:

```
def _calc_empty_plate_const():

    constants.WELL_BG_ARRAY = np.zeros((8, 12, row, col), dtype='float64')
    constants.WELL_INT_ARRAY = np.zeros((8, 12, row, col), dtype='float64')
[...]

PR #70, opened 2020-05-27
Nonetype well registration

bryantChhun:
***BUG FIX***
1) This PR fixes issue #63, which describes a complaint that the `od_well` object is not subscriptable. This occurs when there is a well registration failure, causing the `od_well` to not be written.  Later, when all the `od_wells` are pulled for report generation, the error is thrown.

2) This PR also fixes another bug regarding MockRegionprop -- if the shape of the supplied image is different from the shape of the "disk" mask, it will throw an error.  This occurred in rare cases where the spot centroid is near an image edge, and sometimes other cases.

3) A secondary problem occurred when trying to fix this:
- the written excel spreadsheets would open with a warning from Excel about corrupted data.  It would attempt to fix, but not obviously do anything.
- the cause of this warning is that an antigen worksheet is written with too many characters (>31).

***Solution***
1) The Nonetype problem is solved by including a check at the time of report generation.  Now, when None is found for the well, it will return -999 (rather than attempt to subscript).
2) This PR modifies the way MockRegionprop automatically calculates median intensity.  It now includes a shape check before attempting to slice into the supplied image data.
3) This PR also contains a check on that length and will auto-truncate if necessary.

### DIFF
--- a/array_analyzer/load/report.py
+++ b/array_analyzer/load/report.py
@@ -2,6 +2,7 @@ import array_analyzer.extract.constants as constants
 from copy import deepcopy
 import numpy as np
 import pandas as pd
+import warnings
 
 
 def write_od_to_plate(data, well_name, array_type):
@@ -43,6 +44,7 @@ def write_antigen_report(writer, array_type):
     :return:
     """
     #todo: loops over all antigens for every well.  So if 6x6 arrays of antigens, 6x6x96 calls
+    # is there a more efficent way to do this?
     well_to_image = {v: k for k, v in constants.IMAGE_TO_WELL.items()}
     for antigen_position, antigen in np.ndenumerate(constants.ANTIGEN_ARRAY):
         if antigen == '' or antigen is None:
@@ -65,12 +67,13 @@ def write_antigen_report(writer, array_type):
         # (this function is called once for each OD, INT, BG)
         od_sheet_df = pd.DataFrame(sheet).T
 
+        sheet_name = f'{array_type}_{antigen_position[0]}_{antigen_position[1]}_{antigen}'
+        if len(sheet_name) >= 31:
+            warnings.warn("antigen sheet name is too long, truncating")
+            sheet_name = sheet_name[:31]
+
         od_sheet_df.to_excel(writer,
-                             sheet_name=f''
-                             f'{array_type}_'
-                             f'{antigen_position[0]}_'
-                             f'{antigen_position[1]}_'
-                             f'{antigen}')
+                             sheet_name=sheet_name)
 
 
 def write_to_sheet(sheet_, well_array_, antigen_position_, well_to_image_):
--- a/array_analyzer/load/report.py
+++ b/array_analyzer/load/report.py
@@ -38,6 +38,7 @@ def write_antigen_report(writer, array_type):
     :param array_type: str one of 'od', 'int', 'bg'
     :return:
     """
+    #todo: loops over all antigens for every well.  So if 6x6 arrays of antigens, 6x6x96 calls
     well_to_image = {v: k for k, v in constants.IMAGE_TO_WELL.items()}
     for antigen_position, antigen in np.ndenumerate(constants.ANTIGEN_ARRAY):
         if antigen == '' or antigen is None:
@@ -81,7 +82,6 @@ def write_to_sheet(sheet_, well_array_, antigen_position_, well_to_image_):
     for position, well in np.ndenumerate(well_array_):
         # extract intensity at antigen_position within this well
         if well is None:
-            print(f"WELL IS NONE, position = {position}")
             val = -999
         else:
             val = well[antigen_position_[0], antigen_position_[1]]
--- a/array_analyzer/load/report.py
+++ b/array_analyzer/load/report.py
@@ -5,46 +5,35 @@ import pandas as pd
 import warnings
 
 
-def write_od_to_plate(data, well_name, array_type):
+def write_od_to_plate(data, well_name, well_array):
     """
     Given data from image_parser.compute_od, write the output to an array
         representing its WELL position
     :param data: np.ndarray output from image_parser.compute_od
     :param well_name: str well name
-    :param array_type: str data type
+    :param well_array: np.ndarray of 8x12 format (96 well plate)
     :return:
     """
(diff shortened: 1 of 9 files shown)


array_analyzer/load/report.py
```diff
@@ -48,29 +55,48 @@ def write_antigen_report(writer, array_type):
         sheet = deepcopy(constants.WELL_OUTPUT_TEMPLATE)
         # loop all wells and write OD, INT, BG of this antigen
         if array_type == 'od':
-            for od_position, od_well in np.ndenumerate(constants.WELL_OD_ARRAY):
-                od_val = od_well[antigen_position[0], antigen_position[1]]
-                well_name = well_to_image[(od_position[0] + 1, od_position[1] + 1)]
-                sheet[well_name[0]][int(well_name[1:])] = od_val
+            sheet = write_to_sheet(sheet, constants.WELL_OD_ARRAY, antigen_position, well_to_image)
```
jennyfolkesson: Why not just use the constants.WELL... as an input parameter? Then you don't have to do an if/else statement.

array_analyzer/load/report.py
```diff
@@ -21,6 +22,10 @@ def write_od_to_plate(data, well_name, array_type):
     else:
         raise AttributeError(f"well name {well_name} is not recognized")
 
+    if data is None:
+        print("\tDATA IS NONE FOR THIS WELL")
```
jennyfolkesson: Should this be a warning? Also, if there's no data, what happens with this array of ones?

array_analyzer/load/report.py
```diff
@@ -48,29 +55,48 @@ def write_antigen_report(writer, array_type):
...
-                             f'{antigen_position[0]}_'
-                             f'{antigen_position[1]}_'
-                             f'{antigen}')
+                             sheet_name=sheet_name)
+
+
+def write_to_sheet(sheet_, well_array_, antigen_position_, well_to_image_):
+    """
+
+    :param sheet_:
```
jennyfolkesson: Add types and descriptions

array_analyzer/load/report.py
```diff
@@ -48,29 +55,48 @@ def write_antigen_report(writer, array_type):
...
+    :param well_array_:
+    :param antigen_position_:
+    :param well_to_image_:
+    :return:
+    """
+    # iterate over all 96 well positions
+    for position, well in np.ndenumerate(well_array_):
+        # extract intensity at antigen_position within this well
+        if well is None:
+            val = -999
```
jennyfolkesson: Would it be possible to write Nones instead? It's just easier to distinguish from real data if it's not numbers...

array_analyzer/load/report.py
```diff
@@ -48,29 +55,48 @@ def write_antigen_report(writer, array_type):
...
         else:
             raise AttributeError(f"report array type {array_type} not supported")
 
-        # write the outputs from the above three to a worksheet
+        # write the outputs from ONE of the above three to a worksheet
+        # (this function is called once for each OD, INT, BG)
         od_sheet_df = pd.DataFrame(sheet).T
 
+        sheet_name = f'{array_type}_{antigen_position[0]}_{antigen_position[1]}_{antigen}'
+        if len(sheet_name) >= 31:
```
jennyfolkesson, line 64: 👍

jennyfolkesson (changes requested):
Thanks! I have some minor comments. These changes are important and hopefully we can merge them soon, so we can think about your TODO for a later PR.

array_analyzer/load/report.py
bryantChhun: that's a great idea.  Then `array_type` becomes just constants.WELL_<>_ARRAY and this whole block becomes one call.

array_analyzer/load/report.py
bryantChhun: Should I move this back to the earlier function `write_antigen_report` now that constants.WELL_.. is a parameter?

array_analyzer/load/report.py
bryantChhun: Ah yes, I can try Nones no problem.  Will test it first.

array_analyzer/load/report.py
bryantChhun: and, i could repeat this style in the above function `write_od_to_plate`

array_analyzer/load/report.py
bryantChhun: interesting.  I think this is a remnant from a test I was doing to see if None types was actually the problem.  Basically this condition is never met because this `write_od_to_plate` is called only if data is not None.
Will remove and test

bryantChhun:
@jennyfolkesson please take a look at the new commits.  In particular, let me know what you think about the changes to MockRegionProp -- such as the shape check on image/mask.  For some tests I ran on difficult data, this check definitely gets used (writes '-1' as median intensity).

EDIT: actually, I just noticed in your other PR that you've made this change to MockRegionProp already.  We can switch to yours if it handles this problem appropriately too.  Maybe ignore this PR's version of MockRegionProp

array_analyzer/load/report.py
jennyfolkesson: I think that makes sense if the remaining function only contains a few lines of code.

array_analyzer/workflows/interpolation_wf.py
```diff
@@ -92,9 +92,9 @@ def interp(input_dir, output_dir):
         pd_OD.to_excel(xlwriter_od_well, sheet_name=well_name)
 
         # populate 96-well plate constants with OD, INT, BG arrays
-        report.write_od_to_plate(od_well, well_name, 'od')
-        report.write_od_to_plate(int_well, well_name, 'int')
-        report.write_od_to_plate(bg_well, well_name, 'bg')
+        report.write_od_to_plate(od_well, well_name, constants.WELL_OD_ARRAY)
+        report.write_od_to_plate(int_well, well_name, constants.WELL_INT_ARRAY)
+        report.write_od_to_plate(bg_well, well_name, constants.WELL_BG_ARRAY)
```
jennyfolkesson, line 97: 👍

jennyfolkesson (approved):
Looks good! I made some changes to regionprop in a later PR (spot_intensity_tune). Once this is merged I can rebase that branch and combine the changes that work best.

PR #72, opened 2020-06-05
fixed bounding box bug

jennyfolkesson:
I didn't see any zeros in ODs today but I did detect a bug in regionprops: I hadn't adjusted the spot image and mask to the new bounding box when doing skimage region props for detected mask.

I also updated the debug plot so it's easier (at least for me) to read, and I made it so that only the intensities within a spot mask are plotter, so it's easier to see exactly which intensities determine the median/mean stats.

Both array_fit and array_interp works now.


mattersoflight (commented):
tested with last experiment and registration works well.

mattersoflight:
the code runs well even when the registration fails. Merging...

PR #74, opened 2020-07-15
Rerun wells

jennyfolkesson:
Added feature in which you can rerun some of the wells with different settings.

If rerunning, you need to use the -r flag and specify the full path to the run dir so that results can be written in the same dir.
You also need to create a sheet in your metadata.xlsx sheet named rerun_wells, with a column named 'well_name' with wells to be rerun.

The feature was pretty straight forward, but I had to restructure the report writing to more easily load existing reports and well xlsx files and add the rerun wells to them, hope that's ok.


bryantChhun:
@jennyfolkesson 
I'm a little confused about what this PR accomplishes.  Why do we need this specialized form of well re-running?  It sounds like it is also reading the results from earlier runs?  If there is a problem with the result of a given well, why not fix it before we write the report?

jennyfolkesson:
@bryantChhun You're right that for a rerun the reports from the previous run are read so that we only need to add/alter the values of the failed wells.
The problem which this PR attempts to fix was that for a given plate certain settings (e.g. fiducials_only=FALSE: using all antigen positions to register to spots) works for most but not all wells. Then the user wants to rerun only the failed wells with new parameters (e.g. fiducials_only=TRUE), and using a new metadata xlsx file helps keeping track of which wells were run with which parameters. Sometimes we don't catch failed registrations or automatically know why they failed, which makes it hard to correct them on the fly.

This was the solution Shalin, Janie and I came up with, but happy to hear other suggestions.

tests/conftest.py
```diff
@@ -67,12 +68,13 @@ def create_good_xml(tmp_path_factory):
     with open(os.path.join(str(input_dir), 'temp.xml'), 'w', encoding='utf-8') as temp_xml:
         temp_xml.write(xmltodict.unparse(doc))
 
-    return str(input_dir)
+    return input_dir, output_dir
```
bryantChhun, line 71: nice, i noticed this problem in a different PR.  Thought to make the change in a different branch and submit PR from there, but here is good too!

tests/utils/spot_regionprop_test.py
```diff
@@ -21,68 +21,92 @@ def spot_and_mask():
...
+    # ROIs
     assert prop.image.all() == im_spot.all()
-    assert 29 <= prop.centroid[0] <= 30
-    assert 34 <= prop.centroid[1] <= 35
-    # There are the coordinates for the mask bounding box
-    assert prop.bbox == [18, 23, 42, 47]
     assert prop.mask.all() == mask_spot.all()
-    assert prop.mean_intensity == np.mean(im_spot[mask_spot > 0])
-    assert prop.median_intensity == np.median(im_spot[mask_spot > 0])
     assert prop.masked_image.all() == (im_spot * mask_spot).all()
```
bryantChhun, line 112: I modified random values for each of the tests to check that it fails (and passes).  This all looks great!

tests/conftest.py
bryantChhun, line 71: I noticed this problem in a different PR.  Thought it could be fixed in the tests branch, but fixing it here is good too!

pysero.py
```diff
@@ -59,14 +60,20 @@ def parse_args():
...
+    parser.set_defaults(rerun=False)
     parser.add_argument(
         '-m', '--metadata',
         type=str,
-        choices=['xml', 'xlsx', 'well'],
-        help="specify the file extension for the experiment metadata"
+        default='pysero_output_data_metadata.xlsx',
+        help="specify the file name for the experiment metadata."
+             "Assumed to be in the same directory as images."
     )
```
bryantChhun, line 78: Are we no longer supporting 'xml' or 'well'/ELISA-style data?

'xml' files were generated by SciReader.  We could ask a user to create an 'xlsx' file even when running SciReader images.

The 'well' type metadata was specifically for running traditional elisa experiments-- where a well was saturated with antibody and produces one value, rather than an array of values.

array_analyzer/load/report.py
```diff
@@ -1,68 +1,132 @@
...
-    :return:
-    """
-    if well_name in constants.IMAGE_TO_WELL:
-        (row, col) = constants.IMAGE_TO_WELL[well_name]
-    else:
-        raise AttributeError(f"well name {well_name} is not recognized")
-
-    well_array[row-1, col-1] = data
 
+class ReportWriter:
```
bryantChhun, line 11: Could be good to add a docstring for this class (here or in __init__ ?).

For example, it appears this class translates constants.antigen_array into a dataframe, but also creates dataframes for all the other properties (intensity, bg, od) -- along with some functions to assign values to those.

bryantChhun (commented):
@jennyfolkesson So far it's good and I like it.  But i do think this is larger than just "rerun wells".  I see this PR as accomplishing multiple things:
1. Changes array data structures into pandas dataframes (but leaves the .xml / .xlsx parsed structures alone)
2. Encapsulates data writing into the `Report` class
3. Adds "rerun" functionality by introducing new metadata sheet
4. number 3 requires that metadata parser looks for a specific file name, whereas before it looked for exactly _one_ .xml or .xlsx

And finally, It should be noted that the workflow does not automatically rerun processing (or does it?).  The user needs to define the errant wells in the new .xlsx and run the program again with the flag "rerun".

pysero.py
jennyfolkesson: Good point - I didn't see much support for 'well' so I removed it and then forgot about it, but if the well option still is in use then I'll definitely need to fix that!

array_analyzer/load/report.py
jennyfolkesson: You're right, this class definitely could use a docstring. One point I was grappling with was if the antigen_array translation to dataframe should happen here or in Metadata. What are your thoughts?

smguo:
> Added feature in which you can rerun some of the wells with different settings.
> 
> If rerunning, you need to use the -r flag and specify the full path to the run dir so that results can be written in the same dir.
> You also need to create a sheet in your metadata.xlsx sheet named rerun_wells, with a column named 'well_name' with wells to be rerun.
> 
> The feature was pretty straight forward, but I had to restructure the report writing to more easily load existing reports and well xlsx files and add the rerun wells to them, hope that's ok.

smguo:
Sorry I accidentally closed this PR when I was commenting. Reopened it now.

I noticed 2 issues when I tried to run this branch on the latest datasets that requires different parameters for different wells:

1. Even without '-r' flag, the branch will still check if images of the wells in the 'rerun_wells' sheet are present in the input folder. Is this behavior expected?

2. I got the following error with some wells:

`Traceback (most recent call last):
  File "/Users/syuan-mingguo/python/serology-COVID19/pysero.py", line 171, in <module>
    run_pysero(args)
  File "/Users/syuan-mingguo/python/serology-COVID19/pysero.py", line 159, in run_pysero
    workflow=args.workflow,
  File "/Users/syuan-mingguo/python/serology-COVID19/pysero.py", line 102, in extract_od
    output_dir,
  File "/Users/syuan-mingguo/python/serology-COVID19/array_analyzer/workflows/registration_workflow.py", line 193, in point_registration
    reporter.assign_well_to_plate(well_name, spots_df)
  File "/Users/syuan-mingguo/python/serology-COVID19/array_analyzer/load/report.py", line 103, in assign_well_to_plate
    spots_row['intensity_median'].values[0]
IndexError: index 0 is out of bounds for axis 0 with size 0 `  

It looks like the error was raised because some of the antigen spots are missing in `spots_df`... but not sure why is the case

array_analyzer/extract/metadata.py
```diff
@@ -51,29 +51,24 @@ def __init__(self, input_folder_, output_folder_):
...
             # check that the xlsx file contains necessary worksheets
             sheets = pd.read_excel(self.xlsx_path, sheet_name=None)
             if 'imaging_and_array_parameters' not in sheets.keys():
                 raise IOError("sheet by name 'imaging_and_array_parameters' not present in excel file, aborting")
             if 'antigen_array' not in sheets.keys():
                 raise IOError("sheet by name 'array_antigens' not present in excel file, aborting")
-
+            # Collect well names for rerun, if sheet exists
+            if 'rerun_wells' in sheets.keys():
+                constants.RERUN_WELLS = list(sheets['rerun_wells']['well_name'])
```
smguo: Maybe also check if '-r' is in the input arguments?

array_analyzer/extract/txt_parser.py
```diff
@@ -138,7 +141,7 @@ def create_xlsx_dict(xlsx):
 
     :param dict xlsx: Opened xlsx sheets
     :return list fiduc: Fiducials and control info
-    :return list spots: None.  spots IDs not needed for .xlsx
+sk    :return list spots: None.  spots IDs not needed for .xlsx
     :return list repl: Replicate (antigen)
```
smguo: What's sk?

array_analyzer/transform/array_generation.py
```diff
@@ -172,22 +163,13 @@ def get_spot_intensity(coords, im, background, params, search_range=2):
...
-                centroid=coord,
-            )
-
-        spot_props[grid_id] = spot_prop
-        bg_props[grid_id] = bg_prop
+        spots_df = spots_df.append(spot_prop.spot_dict, ignore_index=True)
+        spot_props[grid_row_idx, grid_col_idx] = spot_prop
 
-    return spot_props, bg_props
+    return spots_df, spot_props
```
smguo, line 166: I like that we're switching to dataframe for data structure. I'm wondering if we can pass all the essential info from `spot_props` to `spots_df` and just return `spots_df`, or is it necessary to keep both?

jennyfolkesson:
@smguo sorry for the confusion; I was debating whether to use a --rerun flag or just have Metadata figure out if it's a rerun based on if there's a 'rerun_wells' sheet, but it looks like I left it somewhere in the middle... Which strategy would you prefer?

The error message looks like it's trying to read a report that isn't there, probably due to the confusion above. I will straighten that out and we'll see if the issue persists. If you want to I can test it if you tell me what directory/metadata.xlsx you were using. Thanks!

smguo:
@jennyfolkesson I'm fine with either! Maybe it would be easier to just let Metadata figure it out since the rerun sheet is required for rerun anyway? Also this branch now allows the users to provide metadata with different names, so the users can just make a copy of metadata in the same folder for rerun.

Here is the pycharm config that gives the error when I run it:

-e
-i
"/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/igg_fiducial"
-o
"/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/Stitched data from multiple pysero outputs/pysero_biotin_fiducial_20200722_1623"
-wf
"array_fit"
-m
"pysero_output_data_metadata.xlsx"
-d
-r

smguo:
I also noticed the rows and columns of the output OD file seem flipped (was 12 X 8 but 8 X 12 now).

jennyfolkesson:
> I also noticed the rows and columns of the output OD file seem flipped (was 12 X 8 but 8 X 12 now).

@smguo that was by design since most sheets use alphabetical columns and numerical rows so it seemed awkward to do it the other way. But I can change it back if it causes problems downstream.

smguo (commented):
> > I also noticed the rows and columns of the output OD file seem flipped (was 12 X 8 but 8 X 12 now).
> 
> @smguo that was by design since most sheets use alphabetical columns and numerical rows so it seemed awkward to do it the other way. But I can change it back if it causes problems downstream.

@jennyfolkesson I see your point. Yes, Excel uses alphabetical columns and numerical rows, which is unfortunately different from the convention for multi-well plates. 

The output was reformatted to 12 X 8 layout to be consistent with the original plate layout in experiments. I think the idea was to make the output sheet more interpretable even without loading into python. Flipping the dimension doesn't affect me because I always convert the output into the long format dataframe in python before doing analysis (actually long format is easier to load into python). Any thoughts  @jrbyrum13 @bryantChhun @mattersoflight ?

array_analyzer/extract/metadata.py
jrbyrum13: Having the output OD file flipped doesn't affect me either, but others who use it with a different interpretation method may be annoyed by having to manually flip the dimension again for their own use. It is unusual to think of well plates with the columns being alphabetical and the rows numerical.

array_analyzer/extract/txt_parser.py
jennyfolkesson: Probably me saving the file without pressing the command key... :-)

array_analyzer/transform/array_generation.py
jennyfolkesson: The spot_props are holding the ROIs which are used for debug plots... We could potentially rewrite to generate debug images, but maybe save it for a later PR?

jennyfolkesson:
I see. I'll flip back the dimensions to agree with well plate conventions.
Thanks for letting me know!

array_analyzer/extract/metadata.py
jennyfolkesson: I see. I'll flip back the dimensions to agree with well plate conventions.
Thanks for letting me know!

jennyfolkesson:
@smguo Did you do an initial run with this branch before doing a rerun? Since I changed the spots information to be a dataframe in this PR it's trying to load a 'stats_per_well.xlsx' from your RUN_PATH and if it can't find it there will be problems.
I've added an assertion when rerunning to make sure this file exists, maybe that helps.

> Here is the pycharm config that gives the error when I run it:
> 
> -e
> -i
> "/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/igg_fiducial"
> -o
> "/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/Stitched data from multiple pysero outputs/pysero_biotin_fiducial_20200722_1623"
> -wf
> "array_fit"
> -m
> "pysero_output_data_metadata.xlsx"
> -d
> -r

smguo:
> @smguo Did you do an initial run with this branch before doing a rerun? Since I changed the spots information to be a dataframe in this PR it's trying to load a 'stats_per_well.xlsx' from your RUN_PATH and if it can't find it there will be problems.
> I've added an assertion when rerunning to make sure this file exists, maybe that helps.
> 
> > Here is the pycharm config that gives the error when I run it:
> > -e
> > -i
> > "/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/igg_fiducial"
> > -o
> > "/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/Stitched data from multiple pysero outputs/pysero_biotin_fiducial_20200722_1623"
> > -wf
> > "array_fit"
> > -m
> > "pysero_output_data_metadata.xlsx"
> > -d
> > -r

@jennyfolkesson I did and the initial run did output a 'stats_per_well.xlsx'. The error wasn't due to missing 'stats_per_well.xlsx' file but was because some antigen spots are missing from the dataframe. I couldn't find the file though when I checked it yesterday (Google file stream has been acting up on my mac), so I ran the PR again after pulling the new commits. The error still persists. Here is the new config I use:

-e
-i
"/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/igg_fiducial"
-o
"/Volumes/GoogleDrive/My Drive/ELISAarrayReader/images_scienion/2020-06-24-17-21-32-COVID_June24_OJassay_plate4_images/Stitched data from multiple pysero outputs/pysero_biotin_fiducial_20200730_1421"
-wf
"array_fit"
-m
"pysero_output_data_metadata.xlsx"
-d
-r

jennyfolkesson:
Thank you, I'll debug and see what I find.

jennyfolkesson:
@smguo I found the issue. Sometimes when you only have three fiducials on a line you get misregistrations that don't get caught because there are multiple solutions for spots that fit to those three fiducials.
In this case, the registration was at a pretty high angle, and the way grid rows and columns were calculated in get_spot_intensity then sometimes assigned the wrong row/col index due to coords with high angles being closer to the wrong index.
https://github.com/czbiohub/pysero/pull/74/files#diff-e6e1f3b9710cebc7189c7149ae5450a1L115

There's no need to calculate grid indices from coordinates so I've fixed this issue by replacing row/col indices with itertools.
Please take a look again and see if this works!

smguo:
@jennyfolkesson The new changes fix the error! Yes, the row and column assignments in `get_spot_intensity` was originally written for interpolation workflow. The assignment assumes the spot array has small rotation angle and will fail when the angle is too large. For grid fitting workflow we don't need to re-assign spot indices.

smguo (approved):
Looks great! The error is fixed and the OD report dimension is now consistent with the plate dimension.

PR #75, opened 2020-08-06
Registration class

jennyfolkesson:
I turned the grid registration functions into a class for improved ease of use and readability.

### DIFF
--- a/array_analyzer/transform/array_generation.py
+++ b/array_analyzer/transform/array_generation.py
@@ -58,7 +58,7 @@ def build_centroid_binary_blocks(cent_list, image_, params_, return_type='region
         return target
 
 
-def get_spot_intensity(coords, im, background, params, search_range=2):
+def get_spot_intensity(coords, im, background, search_range=2):
     """
     Extract signal and background intensity at each spot given the spot coordinate
     with the following steps:
@@ -73,8 +73,6 @@ def get_spot_intensity(coords, im, background, params, search_range=2):
         intensity image of the spots (signals)
     :param background: ndarray
         background image without spots
-    :param params: dict
-        parameters parsed from the metadata
     :param float search_range: Factor of bounding box size in which to search for
         spots. E.g. 2 searches 2 * 2 * bbox width * bbox height
     :return pd.DataFrame spots_df: Dataframe containing metrics for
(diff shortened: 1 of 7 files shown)


jennyfolkesson:
FYI: codcov check failed because coverage has gone down compared to master, but that's only because I removed the tests directory from being counted as covered code.

array_analyzer/transform/array_generation.py
```diff
@@ -84,10 +82,10 @@ def get_spot_intensity(coords, im, background, params, search_range=2):
...
     """
     # values in mm
-    spot_width = params['spot_width']
-    pix_size = params['pixel_size']
-    n_rows = params['rows']
-    n_cols = params['columns']
+    spot_width = constants.params['spot_width']
+    pix_size = constants.params['pixel_size']
+    n_rows = constants.params['rows']
+    n_cols = constants.params['columns']
```
smguo, line 88: 👍

array_analyzer/transform/point_registration.py
```diff
@@ -95,157 +64,230 @@ def icp(source, target, max_iterate=50, matrix_diff=1.):
...
+        grid_cols = grid_cols.flatten()
+        grid_rows = grid_rows.flatten()
+        grid_coords = np.vstack([grid_rows.T, grid_cols.T]).T
+        return grid_coords
+
+    def create_gaussian_particles(self,
+                                  mean_point=(0, 0),
+                                  scale_mean=1.,
+                                  angle_mean=0.):
+        """
```
smguo, line 134: This is a minor point but is there a reason to make the mean of the gaussian a class method input, but the std a constant in `constant`? And it seems to me the default values for mean will always be used? Maybe consider moving the mean to `constant` as well?

tests/transform/point_registration_test.py
```diff
@@ -1,15 +1,37 @@
...
+import array_analyzer.extract.constants as constants
 import array_analyzer.transform.point_registration as registration
 
-im_shape = (50, 75)
-reg_coords = np.ones((5, 2), dtype=np.float32)
 
+@pytest.fixture
+def register_inst():
+    im_shape = (50, 100)
+    spot_coords = np.zeros((10, 2), dtype=np.float32)
```
smguo: If I understand this correctly, you set up the test to register 3 fiducials to 10 spots that are all (0, 0)? Can the fiducials actually register to these spots? Maybe consider setting up the spots in a way that they can be registered by the fiducials, so if some future changes break the registration the test would catch it?

smguo (approved):
I tested the PR and it ran without issues. I only have some minor comments.

array_analyzer/transform/point_registration.py
jennyfolkesson: Good point, I'll move it to constants.

tests/transform/point_registration_test.py
jennyfolkesson: Good suggestion, thanks!

PR #78, opened 2021-01-03
Script to train xgboost trees

smguo:
This PR includes the following changes:
1. Add the script to train xgboost classifiers using information from multiple antigens.
2. Update README with usage of the script and `analyze_od` module.


README.md
```diff
@@ -9,14 +9,9 @@ The project automates estimation of antibody titers from data collected with ELI
...
+## Installation
 
-
-## How to analyze the data?
-
-### Installation
-
-On a typical Winodws, Mac, or Linux computer:
+On a typical Windows, Mac, or Linux computer:
 * Create a conda environment: `conda create --name pysero python=3.7`
```
jennyfolkesson, line 15: Just a suggestions: for simplicity you could also create a conda environment yaml file for creating the environment with a one liner. 
E.g. https://github.com/czbiohub/microDL/blob/master/conda_environment.yml

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+from xgboost.sklearn import XGBClassifier
+import xgboost as xgb
+#%%
+def xgb_fit(model, dtrain, features, target, cv=True, folds=None, cv_folds=5, early_stopping_rounds=20):
+    """Train xgboost model
+    :param object model: XGBClassifier instance
+    :param dataframe dtrain: training data with rows being samples and columns being features
+    :param list features: column names of features
+    :param str target: column name of the target
+    :param bool cv: if true, optimal number of estimators is determined by cross-validation
```
jennyfolkesson: Spell it out as cross_valid or similar, cv is already used for opencv.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    if cv:
+        xgb_param = model.get_xgb_params()
+        xgtrain = xgb.DMatrix(dtrain[features].values, label=dtrain[target].values)
+        cvresult = xgb.cv(xgb_param, xgtrain, num_boost_round=model.get_params()['n_estimators'],
+                          folds=folds, nfold=cv_folds, verbose_eval=True,
+                          metrics='auc', early_stopping_rounds=early_stopping_rounds, stratified=True)
+        model.set_params(n_estimators=cvresult.shape[0])
+        test_score = cvresult['test-auc-mean'].iloc[-1]
+        print('# of estimators: %d' % cvresult.shape[0])
```
jennyfolkesson: Do you want to print this? Either way, add to logger too.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # retrain the model with the whole dataset
+    model.fit(dtrain[features], dtrain[target], verbose=True, eval_metric='auc')
+    train_yhat = model.predict(dtrain[features])
+    train_score = metrics.accuracy_score(dtrain[target].tolist(), train_yhat)
+    if cv:
+        score = test_score
+        print("AUC Score (Test): %f" % score)
```
jennyfolkesson: Add to logger

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    return model, score
+
+def tune_cls_para(model, train, features, target, param_test, cv=None, n_jobs=8):
+    """Tune classifier parameters with sklearn GridSearchCV
+    :param object model: estimator object
+    :param dataframe train: training data with rows being samples and columns being features
+    :param list features: column names of features
+    :param str target: column name of the target
+    :param dict or list of dictionaries param_test: Dictionary with parameters names (string)
+        as keys and lists of
```
jennyfolkesson: Weird line break?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+        parameter settings to try as values, or a list of such
+        dictionaries, in which case the grids spanned by each dictionary
+        in the list are explored. This enables searching over any sequence
+        of parameter settings.
+    :param int, cross-validation generator or an iterable cv: cross-validation splitting strategy.
```
jennyfolkesson: I feel this one could need a bit more of an explanation since there are so many options...

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    :param int or None n_jobs: Number of jobs to run in parallel.
```
jennyfolkesson: None means only one job? Same as 1 or different? Parallel as in multiprocessing?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    :return object model: estimator object with optimal parameters.
+    """
+    gsearch = GridSearchCV(estimator=model, param_grid=param_test,
+                           scoring='roc_auc', n_jobs=n_jobs, iid=False, verbose=3,
+                           cv=cv)
+    gsearch.fit(train[features], train[target])
+    model.set_params(**gsearch.best_params_)
+    means = gsearch.cv_results_['mean_test_score']
+    print(means)
```
jennyfolkesson: Remove print

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    stds = gsearch.cv_results_['std_test_score']
+    for mean, std, params in zip(means, stds, gsearch.cv_results_['params']):
```
jennyfolkesson, line 134: It seems like this function is three lines of code then a bunch of print statements... can some prints be moved to log instead?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+        print("%0.4f (+/-%0.04f) for %r"
+              % (mean, std * 2, params))
+    print('best CV parameters:')
+    print(gsearch.best_params_)
+    print('best CV score: %f' % gsearch.best_score_)
+    return model
+
+def plot_xgb_fscore(model, output_dir, output_fname):
+    """plot xgboost feature importance"""
```
jennyfolkesson: Add params.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
+import pandas as pd
```
jennyfolkesson: the interpretation module needs an __init__.py

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+import os
+from matplotlib import pyplot as plt
+import unicodedata
+from interpretation.plotting import get_roc_df, roc_plot_grid, thr_plot_grid
+from interpretation.report_reader import slice_df, normalize_od, offset_od
+from sklearn import metrics
+from sklearn.model_selection import GridSearchCV, GroupKFold, GroupShuffleSplit
+from xgboost.sklearn import XGBClassifier
+import xgboost as xgb
+#%%
```
jennyfolkesson: For pep you need to add double spacing between functions.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    fig = plt.figure()
+    fig.set_size_inches((6, 6))
+    xgb.plot_importance(model, importance_type='gain')
+    plt.savefig(os.path.join(output_dir, ''.join([output_fname, '.jpg'])),
+                dpi=300, bbox_inches='tight')
+    plt.close()
+
+if __name__ == '__main__':
+    #%% set file paths and load OD table
+    data_dir = r'/Volumes/GoogleDrive/My Drive/ELISAarrayReader/manuscript/data_fig_4/output'
```
jennyfolkesson: This assumes whoever runs it will have the same drive structure and data in it as you... is that what you want?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
+import pandas as pd
+import os
+from matplotlib import pyplot as plt
+import unicodedata
+from interpretation.plotting import get_roc_df, roc_plot_grid, thr_plot_grid
+from interpretation.report_reader import slice_df, normalize_od, offset_od
```
jennyfolkesson, line 27: Ideally imports should be sorted alphabetically with local imports saved for last.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+from sklearn import metrics
+from sklearn.model_selection import GridSearchCV, GroupKFold, GroupShuffleSplit
+from xgboost.sklearn import XGBClassifier
+import xgboost as xgb
```
jennyfolkesson: You've already imported all of xgboost, so call the classifier using xgb.sklearn.XGBClassifier instead of importing again.

jennyfolkesson: You need to add xgboost to requirements.
You also need to make sure that users have OpenMP installed, probably by doing a conda install.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    for col, action, key in zip(slice_cols, slice_actions, slice_keys):
+        df_norm = slice_df(df_norm, action, col, key)
+    #%% Normalize OD and transform the table to wide format for model training
+    df_norm = normalize_od(df_norm, norm_antigen, norm_group)
+    df_norm = offset_od(df_norm, offset_antigen, offset_group)
+    df_norm['antigen_row'] = df_norm['antigen_row'].map(str)
+    df_norm['antigen_col'] = df_norm['antigen_col'].map(str)
+    pysero_df_pivot = df_norm.copy()
+    pysero_df_pivot = pd.pivot_table(pysero_df_pivot, values='OD',
+                                 index=['plate ID', 'well_id', 'serum ID',
```
jennyfolkesson: Weird line breaks - you could define index prior to function call for an easier read.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    plot_xgb_fscore(clf, output_dir=fig_path, output_fname='xgb_feature_importance')
+    antigen_list = ['SARS CoV2 N 50', 'SARS CoV2 RBD 250', 'SARS CoV2 spike 62.5']
+    suffix = '_'.join([pipeline, norm_antigen, 'norm_per_plate', 'mean_rands', str(rand_seed), 'low_c', 'ci'])
+    test_keys = test_keys[['plate ID', 'well_id']].drop_duplicates()
+    roc_df = df_norm.copy()
+    slice_cols = ['serum ID', 'antigen type', 'antigen']
+    slice_keys = [sera_roc_list, ['Diagnostic'], antigen_list]
+    slice_actions = ['drop', 'keep', 'keep']
+    fpr = 0.05
+    ci = 95
```
jennyfolkesson, line 292: Constants should be defined at the top of the file with capitals.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+        as keys and lists of
+        parameter settings to try as values, or a list of such
+        dictionaries, in which case the grids spanned by each dictionary
+        in the list are explored. This enables searching over any sequence
+        of parameter settings.
+    :param int, cross-validation generator or an iterable cv: cross-validation splitting strategy.
+    :param int or None n_jobs: Number of jobs to run in parallel.
+    :return object model: estimator object with optimal parameters.
+    """
+    gsearch = GridSearchCV(estimator=model, param_grid=param_test,
```
jennyfolkesson: Perhaps move each parameter to new line for readability.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+     'colsample_bytree':[i/10.0 for i in range(6,10)]
+    },
+    {
+     'reg_alpha':[0, 1e-5, 1e-2, 0.1, 1, 100]
+    }
+    ]
+    for param_test in param_tests:
+        clf = tune_cls_para(clf, train, features, target='target', param_test=param_test, cv=folds, n_jobs=-1)
+
+    #%% Train the classifier with optimal parameters
```
jennyfolkesson: This comment just confused me because the optimal parameters are from tune_cls_para, or did you find optimal values for learning rate and number of estimators somehow as well?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    for col, action, key in zip(slice_cols, slice_actions, slice_keys):
+        roc_df = slice_df(roc_df, action, col, key)
+    roc_df = pd.merge(test_keys, roc_df, how='left', on=['plate ID', 'well_id'])
+    #%% compute ROC curves and AUC
+    roc_df = roc_df.groupby(['antigen', 'serum ID', 'well_id', 'plate ID',
+                        'serum type', 'serum dilution', 'pipeline', 'secondary ID',
+                        'secondary dilution'])['OD'].mean()
+    roc_df = roc_df.reset_index()
+    roc_df = pd.concat([test, roc_df])
+    _ = roc_plot_grid(roc_df, fig_path, '_'.join(['ROC', suffix]), 'pdf', col_wrap=4, ci=ci, fpr=fpr, hue=hue)
```
jennyfolkesson: Do you ever need to return something from a plot function? Maybe consider dropping the return?

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+    clf.set_params(**param)
+    clf, score = xgb_fit(clf, train, features, target='target', folds=folds, early_stopping_rounds=10000)
+    #%% Model prediction
+    for df in [train, test]:
+        df.loc[:, 'OD'] = clf.predict_proba(df[features])[:, 1]
+        df.loc[:, 'antigen'] = 'combined'
+        df.loc[:, 'antigen type'] = 'Diagnostic'
+    #%% transform the dataframe back to long format for plotting
+    test_keys = test.drop(features + ['target'], axis=1)
+    plot_xgb_fscore(clf, output_dir=fig_path, output_fname='xgb_feature_importance')
```
jennyfolkesson: Maybe add some more comments in this function to make talk the user through what's happening.

interpretation/train_classifier.py
```diff
@@ -0,0 +1,219 @@
...
+def plot_xgb_fscore(model, output_dir, output_fname):
+    """plot xgboost feature importance"""
+    fig = plt.figure()
+    fig.set_size_inches((6, 6))
+    xgb.plot_importance(model, importance_type='gain')
+    plt.savefig(os.path.join(output_dir, ''.join([output_fname, '.jpg'])),
+                dpi=300, bbox_inches='tight')
+    plt.close()
+
+if __name__ == '__main__':
```
jennyfolkesson: You should put most of your code in functions or classes, not in __main__. See e.g. 
https://realpython.com/python-main-function/

jennyfolkesson (changes requested):
I added quite a few comments, but I think the person reviewing will appreciate it if you make their lives easy for them. Happy to do a more thorough review once the script is cleaned up a bit!

interpretation/train_classifier.py
smguo: None is the same as 1 according to [joblib's documentation page](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html):

README.md
smguo: Good suggestion! I added a conda environment yaml file

interpretation/train_classifier.py
smguo: Here I'm trying to be consistent with the scikit-learn classifier parameter convention. opencv is almost never passed as a function argument so I think this should be ok?

interpretation/train_classifier.py
smguo: added to logger

interpretation/train_classifier.py
smguo: Added reference to `GridSearchCV`

interpretation/train_classifier.py
smguo: Added init.py

interpretation/train_classifier.py
smguo: Moved this to `example` within the repo

interpretation/train_classifier.py
smguo: Not in this case because they are intended to be local variable

interpretation/train_classifier.py
smguo: No just lower the learning rate and increase the number of rounds empirically. Clarified this in the comment now.

interpretation/train_classifier.py
smguo: Yes the function also returns a dataframe

smguo: Yes the function also returns a dataframe

interpretation/train_classifier.py
smguo: I moved most of the code to the function `main()` now. This is intended to be more a script demonstrating how to build a classifier for pysero output rather a general utility for training classifier (or at least not at the moment). Perhaps we can refactor this later?

smguo:
@jennyfolkesson Sorry for the long delay in getting to this PR. I was working on other projects. I have addressed most of your comments. One thing I should have noted is this PR is intended to be more a script demonstrating how to build a classifier for pysero output rather a general utility for training classifier. We can do more refactor later if this approach turns out to be useful for other projects too. 

@lenafb It would be great if you can help review this PR too. Let me know if you have any question!

lenafb (approved):
This makes sense to me. One suggestion/clarification: should the name of the env conda_env.yml be multiSero?

interpretation/train_classifier.py
jennyfolkesson: Who knows, you might be ok bug-wise... but since cv is used for a different purpose elsewhere in this repo one could argue that you should be consistent with pysero conventions first and other module conventions second.

jennyfolkesson:
@smguo glad to see that pysero is getting some love again. Just let me know if you want me to contribute in any way.
LGMT

jennyfolkesson (approved):

interpretation/train_classifier.py
smguo: Good point on naming convention consistency throughout the repo. I've changed `cv` to `cross_valid`

PR #79, opened 2021-01-07
Fix tests

jennyfolkesson:
It seemed like an old version of pandas (1.0.2) specified in requirements was causing unit tests to break. Please let me know if you don't want the latest pandas version (1.2.0) I can fix it to something like 1.0.5 instead.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ certifi==2019.11.28
 cycler==0.10.0
 matplotlib==3.1.3
 natsort
-numpy==1.18.1
+numpy>=1.18.1
 opencv-python==3.4.2.17
 openpyxl>=2.6.1
 pandas>=1.0.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ natsort
 numpy==1.18.1
 opencv-python==3.4.2.17
 openpyxl>=2.6.1
-pandas==1.0.2
+pandas>=1.0.2
 pyparsing==2.4.6
 scikit-image>=0.16.2
 scipy==1.1.0
--- a/requirements_docker.txt
+++ b/requirements_docker.txt
@@ -5,7 +5,7 @@ natsort
 numpy
 opencv-python==3.4.2.17
 openpyxl
-pandas>=1.0.2
+pandas>=0.25.0
 pyparsing
 scikit-image
 scipy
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ natsort
 numpy==1.18.1
 opencv-python==3.4.2.17
 openpyxl>=2.6.1
-pandas==1.0.2
+pandas>=1.0.2
```
smguo, line 8: I haven't tried running pysero with version higher than 1.0.2. I have no problem with upgrading pandas version if it doesn't break things...

requirements_docker.txt
```diff
@@ -5,7 +5,7 @@ natsort
 numpy
 opencv-python==3.4.2.17
 openpyxl
-pandas>=0.25.0
+pandas>=1.0.2
```
smguo: This requirement file is actually for running inside a docker container on IBM power 9. I had to use the old pandas version to be able to install all the dependency without conflicts. Does the current test also test running in power 9 environment? If not, maybe we can keep this as it is for now?

smguo (changes requested):
Great to see PR from you @jennyfolkesson! Please see my comments below. Let me know your thoughts.

requirements_docker.txt
mattersoflight: I don't see need to run pysero on IBM power 9, given its simplicity. It will be good to have the repo ready to use on more widely available platforms.

jennyfolkesson: Oops, I just realized I still had my github send me notifications to my biohub mailaddress which doesn't exist anymore! Sorry about that, that's why I haven't been responsive.

I just changed both requirements for consistency, happy to change back the docker requirements to make it work with the IBM servers.

jennyfolkesson:
@smguo please let me know if requirements are ok now. There's some new warnings coming from a mismatch between numpy and sklearn so we can either downgrade the numpy version to 1.16 or we can ignore them for now because I don't think it will create any issues.
https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility

smguo:
@jennyfolkesson sorry for the delay in getting to this! I'm fine with either downgrading numpy or ignoring the warnings. I'll approve the PR.

smguo (approved):

PR #91, opened 2021-10-26
Multisero refactor

lenafb:
I have refactored the multiSero codebase such that README, documentation, module names, and command line all reflect the name change from pysero to multiSero. I am suggesting this merge for naming consistency moving forward.


lenafb:
Note: first commit "pysero refactor part i" incorporates heatmap plotting functions from `heatmap-plotting` branch that existed in my local pysero repo

interpretation/plotting.py
```diff
@@ -444,48 +448,156 @@ def joint_plot(df_ori,
...
+def find_spot_type(hmap,str):
+    #hmap is sliced dilution_df_fit
+    vlp_df = pd.DataFrame()
+    for column in hmap:
+        name = column
+        if name[-3:] == str:
+            vlp_df[f'{name}'] = hmap[column]
+    return vlp_df
+
+def find_rvp_spot_type(hmap,str):
```
smguo: This function looks a lot like `find_spot_type`. You can probably make them a single function by doing something like `if str in name:`. Also you can slice the dataframe by column names rather than create a new dataframe and assign column by column.

interpretation/plotting.py
```diff
@@ -444,48 +448,156 @@ def joint_plot(df_ori,
...
+    vlp_t = vlp_df.transpose()
+    #vlp_t = vlp_t.drop(columns='100-100')
+    fig, ax = plt.subplots(figsize=(x, y))
+    sns.heatmap(vlp_t, annot=True, ax=ax, vmin=vmin, vmax=vmax)
+    plt.xticks(rotation=0)
+    plt.yticks(rotation=0,fontsize=16)
+    plt.title(f'{type} Values per Antigen per Serum ID ({str})', fontsize=20)
+    plt.savefig(os.path.join(fig_path, '.'.join([f'{str}_{type}_map', ext])), dpi=300, bbox_inches='tight')
+
+def delta_ic50(vlp_df,fig_path,ext,index2type,str):
```
smguo: What is this function plotting? Maybe add doctrings?

interpretation/plotting.py
```diff
@@ -444,48 +448,156 @@ def joint_plot(df_ori,
...
+        split_subplots_vals = dilution_df[split_subplots_by].unique()
+        mks = itertools.cycle(['o', 'v', '^', '<', '>', '8', 's', 'p', '*', 'h', 'H', 'D', 'd', 'P', 'X'])
+        plot_by_type(std_by_spot, mks, dilution_df, dilution_df_fit, split_subplots_by, split_subplots_vals, fig_name, fig_path,
+                     ext, hue, col_wrap, zoom=False)
+        fig_name += '_next'
+        spot_df = hmap.filter(regex=y)
+        plot_heatmap(spot_df,fig_path,ext,str=y,type='IC50',vmin=0,vmax=ic_vmax,x=45,y=15)
+        delta_ic50(spot_df,fig_path,ext,index2type,str=y)
+
+    vlp_b_df = bmap.filter(regex='VLP')
```
smguo: You're adding quite a few heatmap plots to this the function so the function now plot both standard curves and 5 heatmaps. I would suggest separate them into two different functions to avoid confusion.

interpretation/plotting.py
```diff
@@ -444,48 +448,156 @@ def joint_plot(df_ori,
...
+    hmap = alt.pivot(index=None,columns='antigen',values='c')
+    bmap = alt.pivot(index=None, columns='antigen', values='b')
+    dmap = alt.pivot(index=None, columns='antigen', values='d')
+
+    hmap.to_csv(fig_path + 'hmap.csv')
+    dmap.to_csv(fig_path + 'dmap.csv')
+
+    slope_vmax = 3
+    ic_vmax = 0.001
+
```
smguo: Try to avoid hardcoding specific sample names in the general plotting function. You can save the serotype in the metadata tab `serum cat` or `serum type` instead of here.

smguo (changes requested):
The PR looks good @lenafb! I left a few comments on the changes to `plotting` module. Let me know if you have any questions!

interpretation/plotting.py
lenafb: I just pushed changes reorganizing these two functions such that if the user is trying to evaluate the performance of individual antigens, they will get both types of plots, but if a user is wants to split subplots by antigen, they will just get  standard curves. Let me know what you think!

lenafb:
Hi @smguo, I've pushed commits addressing your previous comments, I think that this version is a little easier to parse and debug should a user need to tweak parameters.

PR #99, opened 2023-06-04
BSD-3 + requirements

mattersoflight:


requirements.txt
```diff
@@ -1,17 +1,17 @@
-certifi==2019.11.28
-cycler==0.10.0
-matplotlib==3.1.3
+certifi
+cycler
+matplotlib
 natsort
 numpy>=1.18.1
-opencv-python==3.4.2.17
+opencv-python
```
ziw-liu: The other simplifications look good. However I'm not very familiar with OpenCV. Will the later versions have the same behavior?

mattersoflight: good point about OpenCV, should be tested before we make a change. I think the versions were pinned for jetson nano, which no longer applies. Are you currently analyzing on HPC or on M1? I made these changes while setting up env on M1. Can you document your current environment instead?

ziw-liu: I have been working on an M1 laptop with this pinned version. It works fine so I think we can keep this just for consistency.

mattersoflight: pinned now.

mattersoflight:
@ziw-liu I just restored the requirements. Merging now with BSD3 license.

